Issue #23, opened 2020-01-20
Support Changes in Metrics Tags for Prometheus


sklose:
Currently all tags used for metrics need to be pre-registered with the prometheus library. CC metrics API doesn't allow the Prometheus module to know all possible tags ahead of time though.

We need to detect at runtime when new tags are being added and then recreate the Prometheus metric with the superset of all known tags at that point. It is important to ensure that the data that has been recorded, but was not scraped yet, is not lost.

PR #38, opened 2020-01-20
Support Changes in Metrics Tags for Prometheus

plameniv:
Implement a limited feature-set prometheus client library to remove the need to pre-register labels.

The implemented features are the minimum needed to implement the IMetrics interface and to allow for successful scraping by Prometheus of Counter, Gauge and Histogram metrics.

Addresses #23


packages/prometheus/src/config.ts
```diff
@@ -23,25 +23,33 @@ export interface IPrometheusConfiguration {
...
+    /**
+     * Defines a map between keys and histogram buckets to use instead of the default.
+     * All metrics with the same key (regardless of lables) will share the same histogram.
+     *
+     * Defaults to empty map.
+     *
+     * @type {Map<string, number[]>}
+     * @memberof IPrometheusConfiguration
+     */
+    readonly mapOfHistogramBucketsPerKey?: Map<string, number[]>;
```
sklose: maybe name this just `histogramBuckets` ... don't think we need to include a description of the data type in the field name

packages/prometheus/src/models.ts
```diff
@@ -0,0 +1,273 @@
...
+    [key: string]: string | number;
+}
+
+interface ICounter {
+    name: string;
+    count: number;
+    timestamp: number;
+    labelObject: ILabelValues;
+    increment(value: number, timestamp: number): void;
+    toPrometheusString(): string;
```
sklose: maybe this `toPrometheusString` should be an interface by itself since it's repeated in every metric kind? `IPrometheusMetric` maybe?

packages/prometheus/src/index.ts
```diff
@@ -114,33 +136,71 @@ class PrometheusMetrics implements IMetrics, IRequireInitialization, IDisposable
...
 
+    private lookupBuckets(key: string): number[] {
+        const buckets = this.keyBucketsMap.get(key);
+        if (!buckets) {
+            return this.config.defaultHistogramBuckets;
+        }
+        return buckets;
+    }
+
+    private containsLeLabel(tags: IMetricTags): boolean {
```
sklose: what's a LeLabel?

packages/prometheus/src/models.ts
```diff
@@ -0,0 +1,273 @@
...
+        let ii = 0;
+        for (; ii < keys.length - 1; ii++) {
+            str += `${keys[ii]}:${labels[keys[ii]]},`;
+        }
+        str += `${keys[ii]}:${labels[keys[ii]]}`;
+    }
+    return str;
+}
+
+function generateLabelsTag(labelObject: ILabelValues, skipBraces?: boolean): string {
```
sklose: would this generate two different keys for?

```json
{
   "label": undefined
}
```

and 

```json
{
}
```

wondering if this should map to same key

plameniv: When I was testing with the prom-client lib, I noticed that it took labels of the form `{ label1: undefined }` and output a string which had `key{label1="undefined"}. That's how I've modeled our label conversion as well.

This comes from the fact that the allowed types of a label value in Prometheus are `number` and `string`, while we allow tags of `any` type.

The example you gave will produce 2 different label sets. For `key`:
`{ label: undefined }` will produce `key{label="undefined"}`
`{}` will produce `key`.

packages/prometheus/src/index.ts
plameniv: `le` stands for `less than or equal to`. The 'le' label is reserved for histograms so we need to check if someone is trying to use it.

packages/prometheus/src/config.ts
plameniv: How about `histogramBucketsMap`? 
`histogramBuckets` is plural for a single set of buckets and plural for several sets of buckets.

packages/prometheus/src/index.ts
plameniv: I've added a comment above this function explaining its purpose

packages/prometheus/src/models.ts
sklose: shouldn't we strip empty labels though? if I do `metric.increment("x", { "foo": null });` shouldn't that logically be equivalent to doing `metric.increment("x", { });` ?

plameniv: I don't want to introduce another thing which behaves differently from the prom-client library.

packages/prometheus/src/config.ts
sklose: I am not a fan of data types being part of field names ... thinking about it we might actually wanna change the data type to `{ key: string; buckets: numbers[] }[]` and only internally create the map of it ... the end user should't need to care how we store things internally

packages/prometheus/src/models.ts
sklose: well the prom-client only accepts numbers and strings as arguments though ... passing `undefined` or `null` is already undefined behavior since this only works by forcing it with a cast to `any`, right?

plameniv: Since a label set `{}` and a label set `{ label: ""}` both result in `key` with no explicit label information, I've made the change to treat `""`, `undefined`, `null` as the same and just drop labels with such values from the label set

packages/prometheus/src/__test__/prometheus.integration.test.ts
```diff
@@ -0,0 +1,335 @@
...
+            await sleep(500);
+        }
+    }
+    return false;
+}
+
+describe("Prometheus", () => {
+    it("correctly queries Prometheus for several Counters", async () => {
+        const config = { ...defaultConfig, port: port1 };
+        const prom = makeLifecycle(prometheus(config));
```
sklose: can you ensure it initializes itself to DefaultComponent context so we dont have to call these two functions explicitly here?

packages/prometheus/src/__test__/prometheus.integration.test.ts
```diff
@@ -0,0 +1,335 @@
...
+    expect(resp.status).toEqual("success");
+    expect(resp.data.resultType).toEqual("vector");
+}
+
+const prefix = "pre_";
+const key1 = "key1";
+const key2 = "key2";
+const cc = "c_";
+const gg = "g_";
+const hh = "h_";
```
sklose: some of this seems to be used by single test cases only, can you move it to be a local variable there?

packages/prometheus/src/config.ts
```diff
@@ -7,6 +7,11 @@ LICENSE file in the root directory of this source tree.
 
 import { config } from "@walmartlabs/cookie-cutter-core";
 
+export interface IConfiguredHistogramBuckets {
+    key: string;
```
sklose: think this can be readonly

packages/prometheus/src/index.ts
```diff
@@ -114,33 +140,86 @@ class PrometheusMetrics implements IMetrics, IRequireInitialization, IDisposable
...
+            } else if (isNumber(tags[key])) {
+                labels[key] = tags[key];
+            } else if (!isNullOrUndefined(tags[key])) {
+                labels[key] = tags[key].toString();
+                badLabels.push(key);
+            }
+        }
+        if (badLabels.length > 0) {
+            const str = "Prometheus Labels of Type other than string or number passed in";
+            this.logger.warn(str, { listOfLabels: badLabels });
```
sklose: type name in field name again :-) just "labels" is good enough

packages/prometheus/src/__test__/prometheus.integration.test.ts
plameniv: The initialize is needed to start the HttpServer. 

I can make the initialize's argument be optional since I am already constructing the PrometheusMetrics object with `this.logger = DefaultComponentContext.logger;`

I can also inline the config object.

packages/prometheus/src/config.ts
plameniv: I've made both `key` and `buckets` readonly.

packages/prometheus/src/index.ts
plameniv: Renamed to `badLabels`

packages/prometheus/src/__test__/prometheus.integration.test.ts
sklose: nvm then, i forgot about the http server ... leave it as is

plameniv: I can't make the context in the initialize optional because that's how the interface is defined.

packages/prometheus/src/__test__/prometheus.integration.test.ts
plameniv: moved things and refactored a few things

PR #40, opened 2020-01-22
Integration command exits with jest exit code

plameniv:
Travis was reporting success on failed integration tests. This change fixes the issue.

The reason for the failure was that the yarn integrate command:
`docker-compose up -d && jest <...>; docker-compose down`
exits with the exit code from the docker-compose down command instead of the jest one.

The following modified version temporarily stores the exit code from jest and exits with it at the end:
`docker-compose up -d && jest <...>; R=$?; docker-compose down; bash -c "exit $R"`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -771,10 +771,10 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
-"@google-cloud/bigquery@4.1.8":
-  version "4.1.8"
-  resolved "https://registry.yarnpkg.com/@google-cloud/bigquery/-/bigquery-4.1.8.tgz#a2ff3096e00b6d56e0e75ee14603663a41376800"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -771,10 +771,10 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
-"@google-cloud/bigquery@4.1.8":
-  version "4.1.8"
-  resolved "https://registry.yarnpkg.com/@google-cloud/bigquery/-/bigquery-4.1.8.tgz#a2ff3096e00b6d56e0e75ee14603663a41376800"
-  integrity sha512-O6IJlAY4yr35LAz69b238naC2YIjUpiaW2aU4N9OisSQPsBs11YqjkHbgOvxWGbLLPRVCYZZ8ZIpoHjVu4m8gQ==
+"@google-cloud/bigquery@4.6.1":
```
sklose: why did this upgrade to 4.6.1 for you when the package.json says `"@google-cloud/bigquery": "4.1.8"`?

plameniv: The gcp package.json actually has: `"@google-cloud/bigquery": "4.6.1"`

sklose: doh! I was looking at an older checkout

PR #43, opened 2020-01-22
Fix failing Kafka integration test

plameniv:
The `triggers a rebalance after staggering the joining of a second consumer.` integration test was failing because it was expecting the wrong number of message reprocessing to happen.


packages/kafka/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-kafka",
-    "version": "1.2.0-beta.5",
+    "version": "1.2.0-beta.6",
```
sklose: no need to bump

PR #48, opened 2020-01-29
kubernetes: adjust logic for creating a watch

talam:
The k8s watch at times will silently fail so adjust the source to take that into account.

* update kubernetes-client to latest version on master that contains various internal fixes for creation and handling of a watch
* adjust logic for startWatch to wrap watch function inside of a promise and handle timeouts differently

### DIFF
--- a/packages/azure/src/utils/QueueClient.ts
+++ b/packages/azure/src/utils/QueueClient.ts
@@ -390,7 +390,13 @@ export class QueueClient implements IRequireInitialization {
     private getKB(input: string | Buffer) {
         const kb = (n: number) => n / 1024;
         if (typeof input === "string") {
-            return kb(Buffer.byteLength(input, "utf16le"));
+            // QueueClient calculates final kb by creating a buffer from the input and encoding in base64.
+            // Since we don't want to base64 twice to calculate things we can take byte length and multiply
+            // by 8 / 6 to get final number of bytes that would have been output by base64. Every 6 bits of data
+            // is encoded into one base64 character.
+            // https://github.com/Azure/azure-storage-node/blob/0557d02cd2116046db1a2d7fc61a74aa28c8b557/lib/services/queue/queuemessageencoder.js#L76
+            // https://stackoverflow.com/questions/13378815/base64-length-calculation
+            return kb(Math.ceil((Buffer.byteLength(input, "utf8") * 8) / 6));
         }
         return kb(input.byteLength);
     }
--- a/packages/azure/src/utils/QueueClient.ts
+++ b/packages/azure/src/utils/QueueClient.ts
@@ -390,7 +390,7 @@ export class QueueClient implements IRequireInitialization {
     private getKB(input: string | Buffer) {
         const kb = (n: number) => n / 1024;
         if (typeof input === "string") {
-            return kb(Buffer.byteLength(input, "utf16"));
+            return kb(Buffer.byteLength(input, "utf16le"));
         }
         return kb(input.byteLength);
     }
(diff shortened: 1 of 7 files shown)


packages/kubernetes/src/KubernetesWatchSource.ts
```diff
@@ -139,72 +139,92 @@ export class KubernetesWatchSource implements IInputSource, IRequireInitializati
...
-                            result: error ? K8MetricResult.Error : K8MetricResult.Success,
-                        });
-                        if (error) {
-                            failSpan(span, error);
+        const watchPromise = new Promise<string>((resolve, reject) => {
+            watch
+                .watch(
+                    this.queryPath,
+                    this.queryParams,
+                    async (phase: string, obj: any) => {
```
sklose: don't think this can be async based on the API signature ... might have to find a different way

packages/azure/src/utils/QueueClient.ts
```diff
@@ -390,7 +390,7 @@ export class QueueClient implements IRequireInitialization {
     private getKB(input: string | Buffer) {
         const kb = (n: number) => n / 1024;
         if (typeof input === "string") {
-            return kb(Buffer.byteLength(input, "utf16"));
+            return kb(Buffer.byteLength(input, "utf16le"));
```
sklose: why change this?

packages/kubernetes/src/KubernetesWatchSource.ts
talam: changed this around a bit

packages/azure/src/utils/QueueClient.ts
talam: for whatever reason this stopped compiling since `byteLength` takes this in now:
```
type BufferEncoding = "ascii" | "utf8" | "utf-8" | "utf16le" | "ucs2" | "ucs-2" | "base64" | "latin1" | "binary" | "hex";
```

sklose: hmm, this seems to change behavior though since you had to change the assertion to use twice as many bytes. should this be utf8 instead of 16 maybe to match the old byte count?

packages/kubernetes/src/KubernetesWatchSource.ts
```diff
@@ -139,72 +139,111 @@ export class KubernetesWatchSource implements IInputSource, IRequireInitializati
...
-                    const msgRef = new MessageRef({}, msg, span.context());
-                    msgRef.once("released", async (_, __, error) => {
-                        this.metrics.increment(K8Metrics.MsgProcessed, {
-                            event_type: _.payload.type,
-                            result: error ? K8MetricResult.Error : K8MetricResult.Success,
+                        pendingEnqueue.then(() => {
+                            const msgRef = this.createMsgRef(phase, obj);
+                            if (msgRef) {
+                                // tslint:disable-next-line:no-floating-promises
+                                pendingEnqueue = this.queue.enqueue(msgRef);
```
sklose: what if you have two msg coming in while the same element is `pendingEnqueue`? wouldnt the `then` callback be executed twice and one of the two messages will be overwriting the other's assignment to `pendingEnqueue`? wondering if this would be better

```
const msgRef = this.createMsgRef(phase, obj);
if (msgRef) {
    pendingEnqueue = pendingEnqueue.then(() => this.queue.enqueue(msgRef));
}
```

that would chain the enqueues into a linked list

talam: yea, good point. multiple messages coming through could have one getting lost. that approach works better and chains correctly.

packages/azure/src/utils/QueueClient.ts
cross311: This checks UTF16 because that is the format that azure queue takes in the contents.

https://stackoverflow.com/questions/45692731/azure-queue-body-is-too-large-and-exceeds-the-maximum-permissible-limit#comment78346811_45693502

sklose:
hmm, we checked the node implementation though, it does utf8 and then sends that as base64 ... does the SO maybe refer to .net implementation where all strings by default are utf16?

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #51, opened 2020-01-31
kubernetes: fix kubernetes-client package build issue with Node 8

talam:
The `kubernetes-client` package requires that we use a version above Node 8. Since we're currently supporting Node 8 we can't really build and release that package. As a temporary solution till we deprecate Node 8 completely as part of #52 lets adjust the build pipeline to use a separate `package.v8.json` which doesn't build it. Since the package gets built as part of Node 10 and 12 we'll be ok for now. Also, update the build pipeline to not use yarn cache.


.gitignore
```diff
@@ -8,4 +8,10 @@
 
 /docs/website/build
 /yarn-error.log
+/package-lock.json
```
sklose: wanna make this /**/package-lock.json? npm would usually put those files in every package's folder

talam: true, a developer could just npm install in the individual package dir

.travis.yml
```diff
@@ -105,4 +104,19 @@ matrix:
...
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
+  - export PATH="$HOME/.yarn/bin:$PATH"
+
+install:
+  # https://github.com/walmartlabs/cookie-cutter/pull/51
+  # the kubernetes-client package currently doesn't support node 8 so as a workaround
+  # create a special package.json that doesn't try to yarn install it. Rely on other stages
+  # to build and test it.
+  - if node --version | grep -q v8.17.0; then cp package.v8.json package.json; fi
```
sklose: can you make this match v8 more broadly? travis updates their node 8 version once there is a new minor/patch available

sklose: do we need to keep the `cat` and `ls`?

talam: fixed.

talam: missed the `cat` `ls` bit. will address separately.

PR #55, opened 2020-02-03
Allow opt-in to creating Azure queues dynamically on write

chrnola:


packages/azure/src/utils/QueueClient.ts
```diff
@@ -159,22 +160,38 @@ export class QueueClient implements IRequireInitialization {
...
+                QueueMetrics.Write,
+                this.generateMetricTags(queueName, undefined, QueueMetricResults.ErrorTooBig)
+            );
+            throw error;
+        }
+
+        if (this.config.createQueueIfNotExists) {
+            const spanName = "Create Azure Queue (If Not Exists)";
+            const createQueueSpan = this.tracer.startSpan(spanName, { childOf: spanContext });
+            const createQueueIfNotExistsAsync = promisify(this.queueService.createQueueIfNotExists);
```
sklose: are we doing this request on every single write? wondering if we should move this to the `initialize` function and only do it once there.

chrnola: Correct. Not ideal, but what if the set of possible queue names aren't known at the time of initialization?

talam: maybe you can keep a local cache of what you've created and check that on each write.

chrnola: I initially decided against doing that only because queues can be deleted. But in the case where a queue is in the cache but doesn't exist, I think it would be acceptable for the service to crash and let it be created on the next write.

sklose: does azure return maybe a 404 in case the queue doesnt exist? then you could do the logic just based off that and dont even need the map

chrnola: That's a good point, let me look into that.

chrnola: According to [the docs](https://docs.microsoft.com/en-us/rest/api/storageservices/queue-service-error-codes):

| Error Code | HTTP Status Code | User message |
| -- | -- | -- |
| ... | ... | ... |
| QueueNotFound | Not Found (404) | The specified queue does not exist. |

it looks like we're in business. 👍 I'll go ahead and make that change.

cross311:
@chrnola Can you move the queue create into a new function an not inline logic.  We do this in this in other applications where we get the queue via an async function and it tracks if it has seen that name before and it it has not it createIfExists on it, else it just returns.  Make the code much simpler. Since the client is created once for the whole application using an application queue checked already in memory structure should be good. so it only checks once per queue name per app start.

sklose:
@cross311 I think I like @chrnola 's approach better ... that way it's not a new API that we need to make sure we don't break and for the end-user it's just a flag to set vs. manually calling the `createQueue` function and making sure it's called the right amount of time etc. I'd rather keep this an implementation detail behind a config flag than exposing the underlying call to the outside.

packages/azure/src/utils/QueueClient.ts
chrnola: Alright this has been implemented, if we agree (in principle) I'll work on adding test coverage in the morning.

packages/azure/src/utils/QueueClient.ts
```diff
@@ -159,64 +182,79 @@ export class QueueClient implements IRequireInitialization {
...
+                    );
+                    return reject(error);
                 }
-            );
+
+                this.queueService.createMessage(
+                    queueName,
+                    text,
+                    options,
+                    async (
```
sklose: could `async` cause any issues here? the API expects a synchronous callback here I think

packages/azure/src/utils/QueueClient.ts
```diff
@@ -143,6 +146,26 @@ export class QueueClient implements IRequireInitialization {
         span.setTag(QueueOpenTracingTagKeys.QueueName, queueName);
     }
 
+    private async createQueueIfNotExistsAsync(
```
sklose: nit: can you drop the `Async` suffix here?

packages/azure/src/utils/QueueClient.ts
chrnola: Test coverage has been added, let me know what y'all think.

packages/azure/src/utils/QueueClient.ts
chrnola: Good catch, that's a holdover from an earlier take on this. Removed!

packages/azure/src/utils/QueueClient.ts
chrnola: Sure, updated

chrnola:
Is there anything else we should do before merging this? Would like to get a beta build out soon with this change for a new service we're working on.

sklose:
No all good ... forgot that you can't merge yourself. You will want to update the package.json version though if want to see this in a build soonish.

chrnola:
Package version has been updated (good catch).

packages/azure/src/utils/QueueClient.ts
```diff
@@ -143,6 +146,26 @@ export class QueueClient implements IRequireInitialization {
...
+    private async createQueueIfNotExists(
+        spanContext: SpanContext,
+        queueName: string
+    ): Promise<void> {
+        const spanName = "Create Azure Queue (If Not Exists)";
+        const createQueueSpan = this.tracer.startSpan(spanName, { childOf: spanContext });
+        createQueueSpan.log({ queueName });
+
+        try {
+            const createQueueIfNotExistsAsync = promisify(this.queueService.createQueueIfNotExists);
```
cross311: @sklose @chrnola This might end up being an issue in real production code, since you are not binding this to queue service.

`const createQueueIfNotExistsAsync = promisify(this.queueService.createQueueIfNotExists).bind(this.queueService);`

sklose: fixed here https://github.com/walmartlabs/cookie-cutter/pull/63

PR #60, opened 2020-02-06
core: update ConsoleLogger to correctly log out nested objects

talam:
The console logger renders out `[object Object]` if a field is an object so destructure it into `.` delimited nested fields (e.g. foo.bar.baz)

### DIFF
--- a/packages/core/src/__test__/defaults/ConsoleLogger.test.ts
+++ b/packages/core/src/__test__/defaults/ConsoleLogger.test.ts
@@ -15,6 +15,13 @@ describe("ConsoleLogger", () => {
         cc: "core",
         file: "ConsoleLogger.test.ts",
         array: ["one", { two: { foo: "bar" } }],
+        maxDepth: {
+            obj1: {
+                obj2: {
+                    obj3: { obj4: { obj5: { obj6: { obj7: "doesn't get here" } } }, key: "val" },
+                },
+            },
+        },
     };
 
     // tslint:disable:no-console
@@ -37,7 +44,7 @@ describe("ConsoleLogger", () => {
     });
 
     it("logs out info msgs", () => {
-        const logger = new ConsoleLogger();
+        const logger = new ConsoleLogger({ maxDepth: 5 });
         logger.info("info logMsg", logMsg);
         logger.info("info nestedLogMsg", nestedLogMsg);
         const msg1 = capturedLogs[0][0];
@@ -52,7 +59,9 @@ describe("ConsoleLogger", () => {
 
         const msg2 = capturedLogs[1][0];
         let cookie;
-        [, info, msg, foo, cookie, cc, file, arr1, arr2] = msg2.split("|");
+        let maxDepth1;
+        let maxDepth2;
+        [, info, msg, foo, cookie, cc, file, arr1, arr2, maxDepth1, maxDepth2] = msg2.split("|");
         expect(info).toBe("  INFO ");
         expect(msg).toBe(" info nestedLogMsg ");
         expect(foo).toBe(" foo.bar.baz=nested_msg ");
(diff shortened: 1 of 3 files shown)


packages/core/src/defaults/ConsoleLogger.ts
```diff
@@ -24,10 +25,20 @@ export class ConsoleLogger implements ILogger {
         this.log("ERROR", message, data || {}, err);
     }
 
+    private flattenObj(data: any, parentKey: string): string {
```
sklose: can you add unit test for the console logger for this feature?

talam: fixed

packages/core/src/__test__/defaults/ConsoleLogger.test.ts
```diff
@@ -0,0 +1,133 @@
+import { ConsoleLogger } from "../../defaults";
```
sklose: can you move this below the license header?

packages/core/src/__test__/defaults/ConsoleLogger.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    afterEach(() => {
+        capturedLogs = [];
+    });
+
+    afterAll(() => {
+        // tslint:disable:no-console
+        console.log = originalLog;
+    });
+
+    it("info", () => {
```
sklose: nit: might wanna use `test("info")` vs. `it` here since it's not forming a sentence like `it("tests some feature")` usually does

packages/core/src/defaults/ConsoleLogger.ts
cross311: should there be a max depth?

packages/core/src/defaults/ConsoleLogger.ts
```diff
@@ -24,10 +25,20 @@ export class ConsoleLogger implements ILogger {
...
+        return Object.keys(data).reduce((p, k) => {
+            const key = parentKey ? `${parentKey}.${k}` : k;
+            if (_.isObject(data[k])) {
```
cross311: What happens for an array? dont see any test around array, what do we expect to be outputted.

packages/core/src/__test__/defaults/ConsoleLogger.test.ts
```diff
@@ -0,0 +1,133 @@
...
+/*
+Copyright (c) Walmart Inc.
+
+This source code is licensed under the Apache 2.0 license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+describe("ConsoleLogger", () => {
+    const logMsg = { foo: "bar", cc: "core", file: "ConsoleLogger.test.ts" };
+    const nestedLogMsg = {
```
cross311: should we put an array in this message as well just to see what happens or at least have it documented in test what happens?

cross311:
I now this is def a feature people are looking for, for legacy code that is upgraded to use this new functionality we might up our log size greatly.  Obviously people should fix this, but maybe we make sure to call it out in release notes?

packages/core/src/__test__/defaults/ConsoleLogger.test.ts
talam: fixed.

packages/core/src/__test__/defaults/ConsoleLogger.test.ts
talam: adjusted the messages.

packages/core/src/defaults/ConsoleLogger.ts
talam: i think it's fine to not have max depth and let users adjust log messages with that in mind to their liking.  @sklose thoughts?

packages/core/src/defaults/ConsoleLogger.ts
talam: added in a test case

packages/core/src/defaults/ConsoleLogger.ts
sklose: could be useful if ppl try logging out arbitrary objects they receive from somewhere, e.g. 

```
const response = request("http://some-deeply-nested-object.json");
ctx.log("got response", response);
```

talam: added in maxDepth.

PR #61, opened 2020-02-10
kubernetes: create k8sPollSource that periodically polls for resources instead of using a watch

talam:


packages/kubernetes/src/KubernetesPollSource.ts
```diff
@@ -0,0 +1,216 @@
...
+                }
+            }
+
+            this.logger.info(`Finished poll watch`, {
+                queryPath: this.queryPath,
+                queryParams: this.queryParams,
+                reconnectTimeout: this.reconnectTimeout,
+                cachedKeys: this.pollCache.keys(),
+            });
+            await sleep(this.reconnectTimeout);
```
sklose: might have to convert this into something like this:

```
const nextPoll = Date.now() + this.reconnectTimeout();
while (Date.now() < nextPoll && this.isRunning) {
    await sleep(100);
}
```

otherwise it will not terminate gracefully if `reconnectTimeout` is greater than a few seconds

packages/kubernetes/src/KubernetesPollSource.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    public async poll(kubeConfig: k8s.KubeConfig, path: string, queryParams: any): Promise<void> {
```
sklose: you probably don't want to make this `async` and `return new Promise` ... isn't that a Promise returning a Promise 😕

packages/kubernetes/src/KubernetesPollSource.ts
talam: updated

packages/kubernetes/src/KubernetesPollSource.ts
talam: fixed

PR #65, opened 2020-02-18
Add preprocessor concept for Azure Queues to support messages not following the envelope format

butorine:
This PR is to start a conversation around how to allow us to consume any message from the azure storage queue.  This will hopefully resolve 2 issues:
1.  Message text is Base64 encoded and thus fails JSON.parse call.  We need to be able to specify QueueMessageEncoder for QueueService.messageEncoder

2. Messages may not have header and payload, so in this case need a generic header with messageText as the payload.

### DIFF
--- a/packages/azure/package.json
+++ b/packages/azure/package.json
@@ -14,7 +14,6 @@
     },
     "dependencies": {
         "@azure/cosmos": "3.4.2",
-        "@types/node": "^13.7.1",
         "azure-storage": "2.10.3",
         "lodash": "4.17.15",
         "opentracing": "0.14.4",
--- a/packages/azure/package.json
+++ b/packages/azure/package.json
@@ -14,7 +14,6 @@
     },
     "dependencies": {
         "@azure/cosmos": "3.4.2",
-        "@types/jest": "^25.1.2",
         "@types/node": "^13.7.1",
         "azure-storage": "2.10.3",
         "lodash": "4.17.15",
--- a/packages/azure/package.json
+++ b/packages/azure/package.json
@@ -14,6 +14,7 @@
     },
     "dependencies": {
         "@azure/cosmos": "3.4.2",
+        "@types/jest": "^25.1.2",
         "@types/node": "^13.7.1",
         "azure-storage": "2.10.3",
         "lodash": "4.17.15",
--- a/packages/azure/package.json
+++ b/packages/azure/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-azure",
-    "version": "1.2.0-beta.10",
+    "version": "1.2.0-beta.11",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
@@ -14,6 +14,7 @@
     },
     "dependencies": {
         "@azure/cosmos": "3.4.2",
+        "@types/node": "^13.7.1",
         "azure-storage": "2.10.3",
         "lodash": "4.17.15",
         "opentracing": "0.14.4",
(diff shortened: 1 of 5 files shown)


packages/azure/src/streaming/index.ts
```diff
@@ -29,6 +30,7 @@ export interface IQueueConfiguration {
     readonly storageAccount: string;
     readonly storageAccessKey: string;
     readonly queueName: string;
+    readonly queueMessageEncoder?: QueueMessageEncoder;
```
sklose: will need to abstract this out differently ... can't expose underlying types of the azure client lib thru our API ... that makes our API leaky and can cause semver issues down the line

sklose: we could do this the same way we do in the kafka module. there is the concept of a message preprocessor

butorine: Agree.  I'll replicate kafka preprocessor here.

packages/azure/package.json
```diff
@@ -14,6 +14,7 @@
     },
     "dependencies": {
         "@azure/cosmos": "3.4.2",
+        "@types/node": "^13.7.1",
```
sklose: can you remove this? we manage this dependency in the workspace file

packages/azure/src/streaming/index.ts
```diff
@@ -29,6 +29,7 @@ export interface IQueueConfiguration {
     readonly storageAccount: string;
     readonly storageAccessKey: string;
     readonly queueName: string;
+    readonly preprocessor?: IQueueMessagePreprocessor;
```
sklose: can you set a default value for this in the `queueSource` function? default would be a preprocessor implementation that doesn't do anything (null object pattern).

packages/azure/src/utils/QueueClient.ts
```diff
@@ -313,10 +314,25 @@ export class QueueClient implements IRequireInitialization {
                         );
                         resolve(
                             results.map<IQueueMessage>((result) => {
+                                result = this.config.preprocessor.process(result);
                                 const mesageObj = JSON.parse(result.messageText) as {
                                     headers: Record<string, string>;
                                     payload: unknown;
                                 };
+
+                                if (!mesageObj.headers) {
```
sklose: what if we changed the interface of the message preprocessor to be

```
type PreprocessedMessage = {
                                    headers: Record<string, string>;
                                    payload: unknown;
                                };
export interface IQueueMessagePreprocessor {
    process(msg: IQueueMessage): PreprocessedMessage;
}
```

Then we can move the entire JSON.parse and header extraction logic into the default impl and the entire pre-processing can be overwritten.

butorine: I agree about moving JSON.parse logic here.  But I don't think PreprocessedMessage is necessary, since IQueueMessage has headers and payload:
```
export type IQueueMessage = QueueService.QueueMessageResult & {
    headers?: Record<string, string>;
    payload?: unknown;
};
```
However, it seems that headers is something internal to cookie-cutter, so is it necessary to expose them here?  If we use IQueueMessage, then headers are optional, otherwise if we use PreprocessedMessage, can we either remove headers or make them optional?

sklose: makes sense ... we can keep the signature then. we could probably just log an error and skip the message if the EventType header is missing

butorine: Sounds good.  I moved JSON parsing to the preprocessor and skip over messages that do not have EventType header.

packages/azure/src/utils/QueueClient.ts
```diff
@@ -110,6 +113,7 @@ export class QueueClient implements IRequireInitialization {
     public async initialize(context: IComponentContext) {
         this.tracer = context.tracer;
         this.metrics = context.metrics;
+        this.logger = context.logger;
```
sklose: need to add `this.logger = DefaultComponentContext.logger` to the constructor

butorine: Good catch.  Done.  Did the same for QueueInputSource.ts as well.

PR #69, opened 2020-02-22
Fix bug when options are ignored in QueueClient

butorine:
QueueClient ignores the options provided to it and doesn't pass numMessages and visibilityTimeout to the azure queue client.


butorine:
I think options could be simplified by removing IQueueReadOptions and using IQueueSourceConfiguration instead.  There is no way to provide a different queue name for IQueueReadOptions anyway and the queue name from streaming.queueSource() is propagated down to reads.

packages/azure/src/utils/QueueClient.ts
```diff
@@ -278,13 +278,18 @@ export class QueueClient implements IRequireInitialization {
...
         const span = this.tracer.startSpan(this.spanOperationName, { childOf: spanContext });
-        const { queueName, visibilityTimeout, numOfMessages } = Object.assign(
-            {},
-            {
-                queueName: this.defaultQueue,
-            },
-            options || {}
-        );
+
+        let queueName: string = this.defaultQueue;
```
sklose: odd ... the Object.assign is supposed to do exactly the same 😕  ... @cross311 you see anything off the the assign?

butorine: Yes, it's very strange that it doesn't.  My only guess is that something prevents it from doing so after the options go through parsing as in the updated test.

PR #73, opened 2020-03-02
Decorator returning class with empty string name

### LINKED ISSUE #64: Node 12.16.0 builds are broken for core package
[Node 12 builds](https://travis-ci.org/walmartlabs/cookie-cutter/jobs/650025283
) are currently broken for the `core` package. A recent update in Travis CI to `12.16.0` is causing the following errors:
```
 FAIL  src/__test__/config.test.ts
  ● Sections › parses empty object and preserves defaults
    TypeError: Class constructor  cannot be invoked without 'new'
      at Object.descriptor.set (src/config.ts:4044:35)
      at apply (src/config.ts:3935:19)
      at Object.parse (src/config.ts:3973:5)
      at Object.<anonymous> (src/__test__/config.test.ts:373:31)
  ● Sections › applies value provided
    TypeError: Class constructor  cannot be invoked without 'new'
      at Object.descriptor.set (src/config.ts:4044:35)
      at apply (src/config.ts:3935:19)
      at Object.parse (src/config.ts:3973:5)
      at Object.<anonymous> (src/__test__/config.test.ts:380:31)
```

The issue seems to only be happening with `RootConfig` but not with the other classes in that file for some odd reason.

plameniv:
Related TypeScript issue:
https://github.com/microsoft/TypeScript/issues/37157

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -12,7 +12,7 @@ import { IClassType } from "./model";
 export type ValueConvertFn = (val: any) => any;
 
 export function section<T extends new (...args: any[]) => {}>(TConstructor: T) {
-    return class extends TConstructor {
+    const tempClass = class extends TConstructor {
         constructor(...args: any[]) {
             super(args);
 
@@ -33,10 +33,12 @@ export function section<T extends new (...args: any[]) => {}>(TConstructor: T) {
             }
         }
     };
+    Object.defineProperty(tempClass, "name", { value: TConstructor.name });
+    return tempClass;
 }
 
 export function extensible<T extends new (...args: any[]) => {}>(TConstructor: T) {
-    return class extends TConstructor {
+    const tempClass = class extends TConstructor {
         constructor(...args: any[]) {
             super(args);
             if (!Object.getOwnPropertyDescriptor(this, "__extensible")) {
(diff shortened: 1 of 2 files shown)


packages/core/src/config.ts
```diff
@@ -288,6 +292,6 @@ function verifyIsSection(obj: any): obj is ISection {
     throw new Error("unexpected type, are you missing the '@section' decorator?");
 }
 
-function isValueConvertFn(obj: any): obj is ValueConvertFn {
-    return !(isFunction(obj) && obj.__proto__.name.length > 0);
+export function isValueConvertFn(obj: any): obj is ValueConvertFn {
```
sklose: why export this fn?

packages/core/src/config.ts
```diff
@@ -33,10 +33,12 @@ export function section<T extends new (...args: any[]) => {}>(TConstructor: T) {
             }
         }
     };
+    Object.defineProperty(tempClass, "name", { value: TConstructor.name });
+    return tempClass;
 }
 
 export function extensible<T extends new (...args: any[]) => {}>(TConstructor: T) {
-    return class extends TConstructor {
+    const tempClass = class extends TConstructor {
```
sklose: can you add a link to the typescript issue here as a comment?

packages/core/src/config.ts
plameniv: Accidental remnant from testing. Removed the export

packages/core/src/config.ts
plameniv: Added

sklose (approved):

PR #75, opened 2020-03-05
Update Cookie Cutter Dependencies

plameniv:
This updates dependencies to highest available major, minor or patch version which do not cause issues.


packages/kubernetes/package.json
```diff
@@ -20,15 +20,15 @@
...
+        "request": "2.88.2",
+        "request-promise-native": "1.0.8"
     },
     "peerDependencies": {
         "@walmartlabs/cookie-cutter-core": "^1.2.0-beta"
     },
     "devDependencies": {
-        "@types/express": "4.17.1",
-        "@types/request-promise-native": "1.0.16",
+        "@types/express": "4.17.3",
```
sklose: there is no express 4.17.3 for this @types update? seems they would go in lock-step ...

sklose:
are you still updating the remaining dependencies (e.g. kafka) ... or do you wanna do that in a separate PR?

packages/kubernetes/package.json
plameniv: There is no 4.17.3 for express. Last 4 is 4.17.1

plameniv:
I made a commit with updated patch versions for all dependencies that had a patch version upgrade available.

Plan on adding a commit for dependencies that have minor version upgrades.

PR #76, opened 2020-03-05
support multiple concurrent writers for BoundedPriorityQueue

talam:
It's possible to get into a state where we've hit our limit on items
that we can add to the bounded priority queue and a large backlog of
items are accumulated. Once we can begin to process that backlog after
a `whenNotFull` promise resolves, the first item in the backlog will
create a new `whenNotFull` and proceed to await it. Subsequent items in
backlog queue will do the same but overwrite the promise of the first
item leading to a chain of promises for backlogged items that are
unresolvable. This PR ensures that we never overwrite the
`whenNotFull` promise for backlogged queue items and we can resolve
them all eventually.

### DIFF
--- a/packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
+++ b/packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
@@ -89,4 +89,30 @@ describe("BoundedPriorityQueue", () => {
         queue.close();
         await expect(p).resolves.toBe("done");
     });
+
+    it("supports multiple concurrent writers", async () => {
+        const queue = new BoundedPriorityQueue<number>(1);
+
+        const enqueuePromises: Array<Promise<boolean>> = [];
+        for (let i = 1; i < 11; i++) {
+            if (i === 10) {
+                queue
+                    .enqueue(i)
+                    .then(() => {
+                        queue.close();
+                    })
+                    .catch();
+                continue;
+            }
+            enqueuePromises.push(queue.enqueue(i));
+        }
+
+        const buffer = [];
+        for await (const item of queue.iterate()) {
+            buffer.push(item);
+        }
+
+        await Promise.all(enqueuePromises);
+        expect(buffer).toMatchObject([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+    });
 });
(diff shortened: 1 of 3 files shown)


packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
```diff
@@ -89,4 +89,29 @@ describe("BoundedPriorityQueue", () => {
...
+                        queue.close();
+                    });
+                } else {
+                    queue.enqueue(i);
+                }
+            }
+        });
+
+        const buffer = [];
+        await sleep(200);
```
sklose: is the sleep required? technically the iterate should not terminate until the queue was closed, which is after you added the last item

packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
```diff
@@ -89,4 +89,29 @@ describe("BoundedPriorityQueue", () => {
         queue.close();
         await expect(p).resolves.toBe("done");
     });
+
+    it("iterates all contained items when there's a large backlog of items", async () => {
```
sklose: maybe call the test "supports multiple concurrent writers" ... I think that is the underlying feature you are adding with this PR

packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
```diff
@@ -89,4 +89,29 @@ describe("BoundedPriorityQueue", () => {
...
+        const buffer = [];
+        await sleep(200);
+        for await (const item of queue.iterate()) {
+            buffer.push(item);
+        }
+
+        expect(buffer).toMatchObject([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+    });
+
+
```
sklose: nit: got some extra blank lines here

packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
talam: nah, it's not. removed.

packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
talam: fixed.

packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
talam: fixed.

Issue #10, opened 2020-01-15
Add a Way to Inspect 'Invalid' Messages


sklose:
Currently Cookie Cutter prints a log message when a detects an input message that fails validation

```typescript
this.logger.error("received invalid message", result.message, {
    type: msg.payload.type,
});
```

The problem is that there is not enough information in this message to pin point the exact message that failed. We don't want to log out the entire serialized message at this point though, since messages can be large and/or can have JSON serialization issues (see Instana issues serializing error objects).

The proposed solution is to add a new (optional) `invalid` message handler (similar to the `before` and `after`) handlers that is called for all invalid messages. This will allow the application to

* log the payload of the message or any other relevant metadata
* throw an Error and force the application to terminate if in Log And Fail mode (if that's appropriate for the use case)

PR #82, opened 2020-03-31
Way to Inspect 'Invalid' Messages

plameniv:
PR addressing #10

### DIFF
--- a/packages/core/src/__test__/Application.test.ts
+++ b/packages/core/src/__test__/Application.test.ts
@@ -263,7 +263,14 @@ for (const mode of [ParallelismMode.Serial, ParallelismMode.Concurrent, Parallel
         });
 
         it("correctly routes appropriate message through the invalid message handler", async () => {
-            const mockInvalid = jest.fn();
+            const mockInvalid = jest.fn((msg: IMessage, ctx: IDispatchContext) => {
+                const count = (msg.payload as Increment).count;
+                if ((msg.payload as Increment).count === 3) {
+                    ctx.publish(Increment, new Increment(10 * count));
+                } else {
+                    ctx.publish(Increment, new Increment(10 * count + 1));
+                }
+            });
             class MyMessageValidator implements IMessageValidator {
                 public validate(msg: IMessage): IValidateResult {
                     if (msg.payload.count % 2 === 0) {
@@ -281,9 +288,10 @@ for (const mode of [ParallelismMode.Serial, ParallelismMode.Concurrent, Parallel
                 .add(
                     new StaticInputSource([
                         { type: Increment.name, payload: new Increment(2) },
-                        { type: Increment.name, payload: new Increment(3) }, // fails input validation
+                        { type: Increment.name, payload: new Increment(3) }, // fails input validation, passes output validation
                         { type: Increment.name, payload: new Increment(4) },
-                        { type: Increment.name, payload: new Increment(6) }, // fails output validation
+                        { type: Increment.name, payload: new Increment(6) }, // passes input validation, fails output validation
+                        { type: Increment.name, payload: new Increment(9) }, // fails input validatin and fails output validation
                     ])
                 )
                 .done()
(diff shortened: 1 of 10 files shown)


packages/core/src/__test__/Application.test.ts
```diff
@@ -262,6 +262,53 @@ for (const mode of [ParallelismMode.Serial, ParallelismMode.Concurrent, Parallel
...
+                .done()
+                .dispatch({
+                    onIncrement: (msg: Increment, ctx: IDispatchContext) => {
+                        if (msg.count === 6) {
+                            ctx.publish(Increment, new Increment(7));
+                        } else {
+                            ctx.publish(Increment, msg);
+                        }
+                    },
+                    invalid: mockInvalid,
```
sklose, line 298: can you make the `invalid` handler use the context to publish a message as well and assert that that is working

packages/core/src/model/dispatch.ts
```diff
@@ -7,6 +7,13 @@ LICENSE file in the root directory of this source tree.
 
 import { ILogger, IMessage, IMetrics, ITracing, StateRef } from ".";
 import { RetrierContext } from "../utils";
+import { IValidateResult } from "./message";
+
+export class NoInvalidHandlerError extends Error {
```
sklose, line 12: i am not a fan of this new error type ... this kinda becomes part of the public API for everybody who wants to implement a customer message dispatcher ... and throwing an error indicating that something doesnt exist feels odd if it's totally valid that the thing does not exist. can we think of a different way to handle this case?

what would happen if we did `ctx.retry.bail(new ValidationError(...))` in case there is no invalid handler (without any special handling of that error type)?

plameniv: The new error type can easily pass the information that we don't have an Invalid Msg Handler out to the Msg processors but if we don't add special handling for it in the processor then it gets thrown out to the application. 

That behavior would be different from the current behavior when we get an invalid message. Currently an invalid input message is just logged and then skipped over regardless of ErrorHandlingMode.

I had initially started doing something which didn't rely on errors but I am not sure it will solve the API issue.

plameniv: Currently, the path for an invalid message is: fail input validation, skip output validation, increment Invalid Message, log “failed validation”.

We are going to a place where we have the possibility of:

1. fail input validation, skip output validation (old scenario)
2. fail input validation, pass output validation
3. fail input validation, fail output validation due to bad output messages
4. fail input validation, pass output validation due to NO output messages (a invalid handler which does not produce output)

The way I have plan on having the code function is for scenario:

1. increment `ErrInvalidMsg` (old scenario)
2. not increment `ErrInvalidMsg`
3. increment `ErrInvalidMsg`
4. not solved yet (plan is to increment `ErrInvalidMsg`)

For 4. I am planning on modifying the output validation check to see if we had fail input validation and no messages generated for output validation in which case I will increment `ErrInvalidMsg`

The reason for the above is that we have 1 log/metrics point for both input and output validation. For this I need to be able to distinguish between some of those cases and I use an error that only bubbles up partly.

Maybe with the new possible scenarios, we can make the input validation get incremented/logged regardless of the output validation results?
I avoided that solution because it sort of results in double increment/log. But if we split those into “input” and “output” or maybe something along those lines?

packages/core/src/model/dispatch.ts
```diff
@@ -35,7 +36,12 @@ export interface IDispatchContext<TState = any> {
 
 export interface IMessageDispatcher {
     canDispatch(msg: IMessage): boolean;
-    dispatch(msg: IMessage, ctx: IDispatchContext): Promise<any>;
+    hasInvalid(): boolean;
```
sklose, line 39: why do we need `hasInvalid`? thought we didn't need any special handling anymore (be it the prev error type or this function)

plameniv: I need that so we don't record a Success when we get an invalid message and no invalid msg handler.

But maybe I misunderstood what the expected result of that should be? Do we want to record Success in that case?

packages/core/src/__test__/Application.test.ts
```diff
@@ -348,6 +457,73 @@ for (const mode of [ParallelismMode.Serial, ParallelismMode.Concurrent, Parallel
             );
         });
 
+        for (const errorHandlingMode of [
+            ErrorHandlingMode.LogAndContinue,
+            ErrorHandlingMode.LogAndFail,
+        ]) {
+            const modeStr =
```
sklose, line 464: i think there is a better way to stringify the enum ... there should be examples in the same file

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
```diff
@@ -202,19 +202,28 @@ export class ConcurrentMessageProcessor extends BaseMessageProcessor implements
...
-                    if (validate(context, this.validator, this.logger)) {
-                        context.complete();
-                    } else {
-                        context.clear();
-                        super.incrementProcessedMsg(
-                            baseMetricTags,
-                            eventType,
-                            MessageProcessingResults.ErrInvalidMsg
-                        );
+                if (result.success || this.dispatcher.canHandleInvalid()) {
```
sklose, line 205: won't this still blow up if the dispatcher doesn't have the `canHandleInvalid` function?

packages/core/src/__test__/Application.test.ts
plameniv: Fixed

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
plameniv: Appears so, but the compiler didn't catch it as an error. Tried it in Typescript playground and it was caught there, but they use a lot of the strict checks.

Fixed

PR #83, opened 2020-03-31
better throughput in RPC mode while guaranteeing correctness of state

sklose:


packages/core/src/internal/batching/SinkCoordinator.ts
```diff
@@ -90,12 +95,42 @@ export class SinkCoordinator implements IRequireInitialization {
...
+            Array.from(ctx.stored).map((m) => m.state),
+        ]);
+
+        const good = bySequenceNumber.successful;
+        const bad = bySequenceNumber.failed.concat(byEpoch.failed);
+
+        const storeResult = await this.storeTarget.handle(good, retry);
         this.emitMetrics(
             this.stored(storeResult.successful),
             this.stored(storeResult.failed),
```
plameniv, line 111: Shouldn't this emit metrics for the contexts which are part of failed from byEpoch and bySequenceNumber?

packages/core/src/defaults/CachingStateProvider.ts
```diff
@@ -31,15 +31,27 @@ export class CachingStateProvider<TState extends IState<TSnapshot>, TSnapshot>
...
     ) {
+        this.callbacks = new Set();
+        this.callbacksDisabledFor = new Set();
         this.cache = new LRU({
             max: options.maxSize || 1000,
             maxAge: options.maxTTL,
+            noDisposeOnSet: true,
+            dispose: (_, val) => {
+                if (!this.callbacksDisabledFor.has(val.key)) {
+                    for (const cb of this.callbacks.values()) {
```
plameniv, line 50: Wouldn't this loop trigger callbacks for keys other than `val.key` when callbacks are enabled for it?

packages/core/src/defaults/InMemoryMaterializedViewStateOutputSink.ts
```diff
@@ -0,0 +1,65 @@
...
+            if (isStoredMessage(msg)) {
+                additions.push(msg.message.payload);
+                counter.set(msg.state.uniqueId, inc + 1);
+                updates.set(msg.state.key, additions);
+            }
+        }
+
+        for (const entry of updates.entries()) {
+            const existing = this.storage.get(entry[0]);
+            existing.seqNum++;
```
plameniv, line 52: Should be replaced with `existing.seqNum += entry[1].length;`

`entry[1]` is an array of TSnapshot, the current line results in the wrong Sequence Number being returned.

packages/core/src/__test__/Application.test.ts
```diff
@@ -674,6 +674,38 @@ for (const mode of [ParallelismMode.Rpc]) {
...
+                        if (msg.count % 7 === 0) {
+                            await sleep(msg.count * 2 + 100);
+                        }
+                    },
+                },
+                mode,
+                ErrorHandlingMode.LogAndFail
+            );
+
+            expect(streams.get("state-1").data.total).toBe(expected);
```
plameniv, line 706: May want to add `expect(streams.get("state-1").seqNum).toBe(input.length);` so that the correctness of the Sequence Number is asserted.

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
```diff
@@ -292,7 +298,7 @@ export class ConcurrentMessageProcessor extends BaseMessageProcessor implements
                     const sequence = context.source.metadata<number>(
                         EventProcessingMetadata.Sequence
                     );
-                    if (sequence !== reproContext.atSn) {
+                    if (this.shouldSkip(sequence, reproContext.atSn)) {
```
plameniv, line 301: Does this function need a unit test?

packages/core/src/internal/batching/helper.ts
```diff
@@ -94,6 +97,110 @@ export function count<T>(
...
+    for (let i = 0; i < items.length; i++) {
+        const bad = accessor(items[i]).filter(
+            (s) => s.epoch !== undefined && s.epoch < epochs.get(s.key)
+        );
+        if (bad.length > 0) {
+            return {
+                successful: items.slice(0, i),
+                failed: items.slice(i),
+                error: {
+                    error: new SequenceConflictError({
```
plameniv, line 121: Are you trying to make this and the next `SequenceConflictErrors` distinguishable from others? If not, can we make these values closer to what the general pattern is?

packages/core/src/internal/batching/helper.ts
```diff
@@ -94,6 +97,110 @@ export function count<T>(
...
+                lookup.set(state.key, { sn: state.seqNum + 1, seq });
+            } else if (l.seq === seq || l.sn === state.seqNum) {
+                lookup.set(state.key, { sn: l.sn + 1, seq });
+            } else {
+                return {
+                    successful: items.slice(0, i),
+                    failed: items.slice(i),
+                    error: {
+                        error: new SequenceConflictError({
+                            actualSn: -1,
```
plameniv, line 186: Would the following not be more accurate?
`actualSn: l.sn`
`expectedSn: state.seqNum`

packages/core/src/__test__/Application.test.ts
sklose: fixed

packages/core/src/defaults/CachingStateProvider.ts
sklose: i am not 100% if the LRU lib might evict other (expired) items when you do get/del/set ... mainly added this as a precaution so evictions will def trigger the event vs. invalidations don't

packages/core/src/defaults/InMemoryMaterializedViewStateOutputSink.ts
sklose: good catch

packages/core/src/internal/batching/SinkCoordinator.ts
sklose: yup, fixed

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
sklose: don't think we need an extra test for it, it's more of an impl detail ... we have tests for concurrent + rpc mode that assert that the state is correct, which is the main guarantee of both modes and there should also be existing tests that show that RPC mode processes messages in parallel

packages/core/src/internal/batching/helper.ts
sklose: fixed

packages/core/src/internal/batching/helper.ts
sklose: yeah, i was trying to make it distinguishable ... the sn fields don't seem to have much meaning in this case since we didn't even compare any sn's ... we only look at epochs at this point.

PR #86, opened 2020-04-09
Update dependencies for Node 8

plameniv:


plameniv:
Those packages had thrown errors or warnings during the `yarn` phase. But after redoing the command, it appears only the semver 7.1.3 version is incompatible with Node 8.

PR #89, opened 2020-04-13
Reveal az blob & queue service url params, to allow for pointing at a local emulator. 

softdev87:


sklose:
can you bump the version number of the azure package as well? otherwise this won't publish a new package.

PR #91, opened 2020-04-14
Better Log Message when a Message Fails to Process and fix to upsertSproc SeqConErr details

plameniv:


packages/core/src/internal/processor/BaseMessageProcessor.ts
```diff
@@ -165,6 +165,7 @@ export abstract class BaseMessageProcessor implements IRequireInitialization {
                 return val;
             } catch (e) {
                 this.logger.error("failed to dispatch message", e, {
+                    ...msg.getAllMetadata(),
```
sklose, line 168: i think this is logging too much, there is some kind of filter on metadata for the msg handler

plameniv: Should I just use a local MetadataLoggerDecorator here? I can also just copy the filter function here as well.

sklose: i'd just do `context.loggger.error(...)` instead of `this.logger.error` ... the Dispatch Context already has the right logger you want.

plameniv: fixed

PR #93, opened 2020-04-16
Add Dead Letter Queue to QueueInputSource

plameniv:


packages/azure/src/streaming/index.ts
```diff
@@ -34,6 +34,8 @@ export interface IQueueConfiguration {
     readonly storageAccount: string;
     readonly storageAccessKey: string;
     readonly queueName: string;
+    readonly deadLetterQueueName?: string;
```
sklose, line 37: let's find some better naming here ... `dequeueCount` is a little odd as top-level config field. maybe nest both of the fields together under an optional `deadLetterQueue: { queueName: string; maxDequeueCount: number }` field?

plameniv: Sounds good, and will allow us to extend the `deadLetterQueue` to have independent configs from the regular queue.

packages/azure/src/streaming/index.ts
```diff
@@ -29,11 +29,21 @@ export interface IQueueMessagePreprocessor {
     process(payload: string): IQueueMessage;
 }
 
+export interface IDeadLetterQueueConfiguration {
+    readonly queueName: string;
+    readonly maxDequeueCount: number;
+    readonly visibilityTimeout?: number;
+    readonly messageTimeToLive?: number;
+    readonly retryCount?: number;
```
sklose, line 37: what are the retryCount and retryInterval being used for here?

plameniv: A config with a dead letter queue, spawns 2 QueueClients, so I made the retryCount and retryInterval independent, in case the user needs to specify different values.

packages/azure/src/streaming/internal/QueueInputSource.ts
```diff
@@ -57,13 +58,32 @@ export class QueueInputSource implements IInputSource, IRequireInitialization {
...
+            const deadLetterConfig: IQueueConfiguration = {
+                createQueueIfNotExists: config.createQueueIfNotExists,
+                encoder: config.encoder,
+                queueName: config.deadLetterQueue.queueName,
+                storageAccessKey: config.storageAccessKey,
+                storageAccount: config.storageAccount,
+                deadLetterQueue: undefined,
+                largeItemBlobContainer: config.largeItemBlobContainer,
+                preprocessor: config.preprocessor,
+                retryCount: config.deadLetterQueue.retryCount,
```
sklose, line 71: you probably wanna default this to the regular queue's config, if the dead letter queues doesnt have an explicit value then

plameniv: Done

packages/azure/src/streaming/internal/QueueOutputSink.ts
```diff
@@ -41,8 +41,16 @@ export class QueueOutputSink implements IOutputSink<IPublishedMessage>, IRequire
...
         for (const msg of output) {
             const queueName = msg.metadata[QueueMetadata.QueueName];
-            const visibilityTimeout = msg.metadata[QueueMetadata.VisibilityTimeout];
-            const messageTimeToLive = msg.metadata[QueueMetadata.TimeToLive];
+            const visibilityTimeoutSeconds = msg.metadata[QueueMetadata.VisibilityTimeout];
+            const messageTimeToLiveSeconds = msg.metadata[QueueMetadata.TimeToLive];
+            const visibilityTimeoutMs = msg.metadata[QueueMetadata.VisibilityTimeoutMs];
+            const messageTimeToLiveMs = msg.metadata[QueueMetadata.TimeToLiveMs];
+            const visibilityTimeout = visibilityTimeoutMs
+                ? Math.floor(visibilityTimeoutMs / 1000)
```
sklose, line 49: should we ceil or round this? if I put 500ms, then it gets translated to 0 vs. 1s would probably be a better value

packages/azure/src/utils/QueueClient.ts
```diff
@@ -360,17 +367,18 @@ export class QueueClient implements IRequireInitialization {
...
+                                messageObj.headers[QueueMetadata.DequeueCount] = (
                                     result.dequeueCount || 1
                                 ).toString();
-                                mesageObj.headers[QueueMetadata.QueueName] = result.queue;
-                                mesageObj.headers[QueueMetadata.TimeToLive] = result.expirationTime;
-                                mesageObj.headers[QueueMetadata.VisibilityTimeout] =
+                                messageObj.headers[QueueMetadata.QueueName] = result.queue;
+                                messageObj.headers[QueueMetadata.TimeToLive] =
+                                    result.expirationTime;
+                                messageObj.headers[QueueMetadata.VisibilityTimeout] =
```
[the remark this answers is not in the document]
plameniv: The data returned here is actually a date and not a time duration in milli/seconds: `"Tue, 21 Apr 2020 16:33:23 GMT"`

I have an issue opened for this: https://github.com/walmartlabs/cookie-cutter/issues/97 

I did not include it in this PR because it may break users who expect the date.

packages/azure/src/streaming/internal/QueueOutputSink.ts
plameniv: I will add `ceil`

plameniv: On second thought, the reason why I added floor was because that's how the config code deals with converting time. It uses floor. If we make it `ceil` there will be mismatch between how the visibility timeout behaves with the same input in two spots in the code.

docs/docs/Module_Azure.md
```diff
@@ -305,3 +305,50 @@ Application.create()
 It is recommended to run the service in Serial mode with `queueSource` because once the message is received from Azure Queues its visibility timeout window starts and running the service in serial mode will decrease the chance of hitting the window timeout as messages are queued up internally in Cookie Cutter in Concurrent mode.
 
 Queues items will be reprocessed if you throw an error in the message handler function. The `DequeueCount` metadata can be used to detect reprocessed messages and skip over those if appropriate.
+
+### Dead Letter Queue
+
+It is now possible to designate a queue to serve as a dead letter queue. `maxDequeueCount` specifies how many times a message can be dequeued before it is sent to the dead letter queue. The visibility timeout and message time to live can also be specified.
```
sklose, line 311: nit: remove the word `now` for the docs, so it doesn't sound like release notes. also maybe clarify the last sentence a bit `The visibility timeout and message time to live will default to the values of the main queue unless the values are explicitly overwritten`

packages/azure/src/streaming/internal/QueueOutputSink.ts
sklose: ok that makes sense ... let's keep floor then

packages/azure/src/utils/QueueClient.ts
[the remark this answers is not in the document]
sklose: k, makes sense, let's deal with this in a separate PR

docs/docs/Module_Azure.md
plameniv: Done.

Issue #52, opened 2020-01-31
Migration to Yarn 2 and Enforcing Node >= 10


talam:
Yarn 2 was [recently released](https://dev.to/arcanis/introducing-yarn-2-4eh1) with various updates that we would benefit from. Let's migrate over to it as part of CC 1.13. Yarn 2 also doesn't support Node 8 anymore so we can use the opportunity to remove Node 8 support from CC.

cc @sklose

sklose:
@emmalynch is this only blocked for going from 1.21 -> 1.22 or also for upgrading all the way to v2?

If it's completely blocked, then let's limit the scope of the PR you already started to just remove node 8 support + add the `--frozen-lockfile` flag ... we can do the yarn upgrade later then once that ticket is resolved.

emmalynch:
I was able to work around that issue thankfully. 

However, there is an outstanding issue with docusaurus, which will probably require us to upgrade that to v2 before we can use pnp though: https://github.com/yarnpkg/berry/issues/818#issuecomment-580871488

For the moment, I've gone ahead and switched to the node-modules linker, so at least we can upgrade to v2 & remove support for node 8.

PR #107, opened 2020-04-27
[blocked] Migrate to Yarn 2 (#52)

emmalynch:
* migrate to yarn 2
 * remove support for node 8
 * enforce node 10 as minimum version


emmalynch:
Blocked by this issue:  https://github.com/yarnpkg/yarn/issues/7890

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "sourceMap": true,
     "module": "commonjs",
     "target": "es6",
-    "lib": ["es2017", "esnext"],
+    "lib": ["es2017", "esnext", "dom"],
```
sklose, line 7: why do we need the dependency on `dom`?

packages/lightstep/package.json
```diff
@@ -20,10 +20,14 @@
         "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
     },
     "devDependencies": {
-        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
+        "@types/jest": "25.1.3",
```
sklose, line 23: i am not a fan of adding all these global dependencies to each package ... that was previously nicely encapsulated in the workspace file in the root. is there a way to do the same thing with yarn 2?

emmalynch: My preference wouldn't be for this either.

I'm new to yarn, so I took the path of least resistance and added them to each dependency. There was no obvious way to avoid this, but I'll verify and update if there's a neater way to do it.

tsconfig.json
emmalynch: It seems as though yarn requires it. It has a dependency on emscripten via libzip which causes an error if we remove dom.

packages/lightstep/package.json
sklose: another route we might take: disable pnp in yarn so it uses the old node_modules way of doing things. we can investigate pnp separately since that seems like a bigger change

emmalynch: Makes sense, most of the issues definitely stemmed from pnp compatibility. I'll update this pr today

tsconfig.json
emmalynch: This isn't needed when using the node-modules nodeLinker setting, so I've removed it

.travis.yml
```diff
@@ -103,19 +99,15 @@ matrix:
                   git config --global user.name "${GH_NAME}"
                   git config --global user.email "${GH_EMAIL}"
                   echo "machine github.com login ${GH_NAME} password ${GH_TOKEN}" > ~/.netrc
-                  cd docs/website && yarn install && GIT_USER="${GH_NAME}" yarn run publish-gh-pages
+                  cd docs/website && yarn install && GIT_USER="${GH_NAME}" yarn docs/website run publish-gh-pages
 
 
 before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.22.4
```
sklose, line 106: can we install berry directly instead of installing 1.22 first and then switching to berry?

.travis.yml
```diff
@@ -103,19 +99,15 @@ matrix:
...
   - export PATH="$HOME/.yarn/bin:$PATH"
+  - yarn set version berry
 
 install:
-  # https://github.com/walmartlabs/cookie-cutter/pull/51
-  # the kubernetes-client package currently doesn't support node 8 so as a workaround
-  # create a special package.json that doesn't try to yarn install it. Rely on other stages
-  # to build and test it.
-  - if node --version | grep -q v8\.* ; then cp package.v8.json package.json && echo copied package.v8.json into package.json ; fi
   - yarn install
```
sklose, line 111: i think we wanted to add the `--frozen-something-something` flag here - unless that's not relevant anymore with yarn v2

package.json
```diff
@@ -6,27 +6,33 @@
     "docs/website",
     "examples/*"
   ],
+  "engines": {
+    "node": ">=0.10.0"
```
sklose, line 10: i think this should be 10.0 ... also do we need to add this to every single package since the workspace itself is not going to be published as a package?

packages/core/package.json
```diff
@@ -20,13 +20,16 @@
         "wtfnode": "0.8.1"
     },
     "devDependencies": {
+        "@types/jest": "25.1.3",
```
sklose, line 23: what was the issue with not adding these dev dependencies? it looks a little random, the jest types are only added to core, typescript is added to all of them, node types are added to some packages ... i would really prefer if we can keep all of these confined in the workspace file, otherwise it becomes a mess keeping the versions aligned

packages/core/package.json
```diff
@@ -20,13 +20,16 @@
...
         "@types/ms": "0.7.31",
-        "@types/wtfnode": "0.7.0"
+        "@types/node": "13.13.4",
+        "@types/wtfnode": "0.7.0",
+        "typescript": "^3.8.3"
     },
     "scripts": {
         "build": "tsc",
-        "test": "jest --config=../../jest.unit.config.js --rootDir=.",
+        "test": "jest --config=../../jest.unit.config.js --rootDir=. ",
```
sklose, line 32: nit: don't think we need that extra space here?

.travis.yml
emmalynch: The `enableImmutableInstalls` setting in the .yarnrc.yml supersedes --frozen-lockfile.

.travis.yml
emmalynch: There's a couple of bugs that can happen when trying to install berry from earlier versions of yarn. The general advice seems to be to install 1.22 & then update to berry.

package.json
emmalynch: Fixed the version.

I double checked the behaviour, it's not needed in individual packages.

packages/core/package.json
emmalynch: The jest & node type dependencies were added before switching to node-modules for the linker, I've removed them now.

For the typescript one - the individual packages will no longer build if we don't specify it in each package unfortunately.

To get around the version alignment problem, it sounds like the resolution field should help: https://yarnpkg.com/configuration/manifest#resolutions. I'll add that in.

packages/core/package.json
emmalynch: fixed!

packages/core/package.json
sklose: https://yarnpkg.com/features/workspaces#workspace-ranges-workspace it sounds like you can just put `workspace:*` as the version number and then it picks it up from the workspace

.travis.yml
sklose: don't think we can set the `enableImmutableInstalls` in the rc file though ... we want to allow ppl to change dependencies locally, just in CI it should be an error if the lockfile is being changed

emmalynch: Ah, makes sense. Updated!

packages/core/package.json
emmalynch: Done

packages/statsd/package.json
```diff
@@ -19,7 +19,7 @@
         "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
     },
     "devDependencies": {
-        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
```
sklose, line 22: why remove this dev dependency?

packages/azure/package.json
```diff
@@ -24,11 +24,12 @@
         "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
     },
     "devDependencies": {
-        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
+        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta.1",
```
sklose, line 27: don't think you want the `.1` here after `beta`

examples/kafka-events/package.json
```diff
@@ -7,6 +7,9 @@
         "@walmartlabs/cookie-cutter-core": "^1.3.0-beta",
         "@walmartlabs/cookie-cutter-kafka": "^1.3.0-beta"
     },
+    "devDependencies": {
+        "typescript": "workspace:*"
```
sklose, line 11: can we do the same `workspace:*` thing with `opentracing`? i think we might already have a bit of a version mix across the packages for that, this seems like a good way to fix that

packages/azure/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-azure",
-    "version": "1.3.0-beta.1",
```
emmalynch, line 3: @sklose do I need to manually update versions in each package, or is there an automated process for that?

examples/kafka-events/package.json
emmalynch: done!

packages/statsd/package.json
emmalynch: It was done mistakenly. I've added it back in & updated other places where it was removed too.

packages/azure/package.json
sklose: i think manual i this case ... we do have a script in the workspace file to set the version for everything and also update the peer/dev deps accordingly ... but that is mainly meant to go from 1.2 -> 1.3 etc (it sets the exact same version for all pacakges)

emmalynch:
This will need to wait for the following yarn issue to be resolved: https://github.com/yarnpkg/berry/issues/1351

.travis.yml
```diff
@@ -114,15 +110,11 @@ matrix:
 
 
 before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.22.4
   - export PATH="$HOME/.yarn/bin:$PATH"
+  - yarn set version berry
```
merceyz, line 115: Running this here is a bad idea, you risk having the CI behave differently than your local version. You should (and I see you have) check in the version of yarn you want to use.

emmalynch: Good point, thanks. I've updated it!

merceyz: You don't have to run the command at all, the correct version is already installed and configured in your `.yarnrc.yml` file

PR #114, opened 2020-05-14
core: unable to close bounded priority queue

talam:
core: fix issue with bounded priority queue not correctly supporting closing of queues
    immediately after it was drained leading to potential `Cannot read
    property 'resolve' of undefined` errors due to whenNotFull being
    undefined.

### DIFF
--- a/packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
+++ b/packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
@@ -115,4 +115,13 @@ describe("BoundedPriorityQueue", () => {
         await Promise.all(enqueuePromises);
         expect(buffer).toMatchObject([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
     });
+
+    it("closes queue immediately after it becomes empty", async () => {
+        const queue = new BoundedPriorityQueue<number>(1);
+        const enqueuePromise = queue.enqueue(1);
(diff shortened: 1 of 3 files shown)


packages/core/src/__test__/utils/BoundedPriorityQueue.test.ts
```diff
@@ -115,4 +115,11 @@ describe("BoundedPriorityQueue", () => {
         await Promise.all(enqueuePromises);
         expect(buffer).toMatchObject([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
     });
+
+    it("closes queue immediately after it becomes empty", async () => {
+        const queue = new BoundedPriorityQueue<number>(1);
+        queue.enqueue(1);
```
sklose, line 121: do we need to eventually await these promises somewhere?

talam: I think for the purposes of this test it's ok to not await them as I am really just trying to ensure that things get closed properly.

sklose: you wanna make sure though that your test finishes gracefully and jest doesnt end up detecting dangling promises ... also would be a good part of the test that those promises will actually resolve and dont hang indefinitely

talam: fixed.

Issue #81, opened 2020-03-20
Support accessing multiple Cosmos Collection within same service


sklose:
Currently the Cosmos State Provider and all Cosmos Sinks are limited to a specific Database/Collection in Cosmos. We should support a single client accessing data from different collections. Proposed syntax for that:

```
const stateRef = await ctx.state.get("@collection/key");
ctx.store(EventType, stateRef, new EventType());
```

If the `@collection` is left out, it will default to the collection specified in the sink's configuration.
Open question: what would be the syntax for a different data base?

Issue #90, opened 2020-04-13
Add Integration Tests for Azure Package


sklose:
Microsoft offers a simulator for Cosmos DB, Blob Storage and Azure Queues that runs on Windows. We should investigate to use that to run full integration tests for the Azure package.

An example on how to setup that emulator can be found here https://github.com/softdev87/cosmos-db-emulator-vm.

plameniv:
Closed by #204

PR #117, opened 2020-05-18
multi cosmos collections (#81)

emmalynch:
Adds support for multiple cosmos collections
 - keys must be provided in the format  "@collection/key" or "key"
 - collectionIds are optional

### DIFF
--- a/packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
+++ b/packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
@@ -20,9 +20,17 @@ const url = "https://carrot.documents.azure.com:443/";
 const key = process.env.COSMOS_SECRET_KEY;
 const databaseId = "tanvir-test-occ";
 const collectionId = "occ";
+const newCollectedId = "occ2";
 const encoder = new JsonMessageEncoder();
 const sink = new CosmosOutputSink({ url, key, databaseId, collectionId, encoder });
 const client = new CosmosClient({ url, key, databaseId, collectionId, encoder });
+const newSink = new CosmosOutputSink({
+    url,
+    key,
+    databaseId,
+    collectionId: newCollectedId,
+    encoder,
+});
 
 describe.skip("Materialized Views", () => {
     const spanContext = {};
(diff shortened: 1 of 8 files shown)


packages/azure/src/utils/helpers.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns a collectionId from a provided key. If the key is not in the expected format,
+ * undefined is returned
+ * @param key A key in the format @collection/partitionKey
+ */
+export function getCollectionInfo(key: string): [string, string] {
```
sklose, line 6: i think this parsing logic is a little too lenient ... it allows for `collection/key` without the @.
maybe let's also change the return type to `{ collectionId: string?; key: string }` since both types in the tuple are string and it's easy to mix up the indices ... with the explicit type the caller can destructure like this `const { collectionId, key } = getCollectionInfo(key);` and there can't be any confusion whether the first or the seconds is the collectionId.

packages/azure/src/__test__/utils/helpers.test.ts
```diff
@@ -0,0 +1,19 @@
+import { getCollectionInfo } from "../../utils/helpers";
+
+describe("Collection Info helper test", () => {
+    it("correctly parses a collectionId and partition key", () => {
```
sklose, line 4: nit: drop the word 'correctly' ... that's kinda true for all unit tests :-) `it("parses a state key with a collectionId overwrite")` and `it("parses a state key without a collectionId overwrite")`.

packages/azure/src/__test__/utils/helpers.test.ts
```diff
@@ -0,0 +1,19 @@
+import { getCollectionInfo } from "../../utils/helpers";
+
+describe("Collection Info helper test", () => {
```
sklose, line 3: `describe("State Key Parsing")`

packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
```diff
@@ -137,6 +145,93 @@ describe.skip("Materialized Views", () => {
...
+                        },
+                        spanContext,
+                        original: new MessageRef({}, null),
+                    },
+                ]),
+                undefined
+            );
+
+            const state = new CosmosStateProvider(DummyState, client, new JsonMessageEncoder());
+            const stateRef = await state.get(undefined, `unknown/${streamId}`);
```
sklose, line 199: this should actually not work, since `unknown` does not have the `@` prefix

packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
```diff
@@ -137,6 +145,93 @@ describe.skip("Materialized Views", () => {
                 seqNum: 0,
             });
         });
+
+        it("retrieves a document in the default collection", async () => {
+            const currentTime = new Date();
```
sklose, line 150: these tests look a little weird ... can we use a full CC Application to populate data in the backend instead of directly using the sink? that is going to be less brittle if we do some internal refactoring, kinda like this:

```
// setup cosmos state
await Application.create()
    .input()
        .add(new StaticInputSource(...))
        .done()
    .dispatch({
        // code to store input as ouput
    })
    .output()
        .stored(cosmosSink( ... ))
    .done()
    .run();

// perform actual test operation
await Application.create()
    .input()
        .add(new StaticInputSource(...))
        .done()
    .dispatch({
        // code to use state provider to read state back
        // back in and store in a list/map or something
    })
    .done()
    .run();

// assert correctness
assert.equals(map/list[index/key], expected);
```

emmalynch: Makes sense, I'll work on that now

packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
emmalynch: I've updated the state key parser, so this case should be fixed. I'll also refactor this test as part of your above suggestion.

packages/azure/src/utils/helpers.ts
emmalynch: I really like this change, thanks. I've updated it now to be more strict on the format too & added an extra unit test specifically for the missing "@" case.

packages/azure/src/utils/helpers.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param inputKey A state key in the format @collectionId/partitionKey
+ */
+export function getCollectionInfo(key: string): { collectionId?: string; partitionKey: string } {
+    const collectionInfo: string[] = key.split("/");
+
+    let collectionId: string;
+    let partitionKey: string;
+
+    if (collectionInfo.length === 1 && !collectionInfo[0].startsWith("@")) {
```
sklose, line 17: i think this filters our valid partitionKeys ... `@my-stream` is valid in cosmos

emmalynch: My mistake, I didn't realise that was a valid format. I've corrected it now.

packages/azure/src/__test__/materialized/internal/MaterializedView.integration.test.ts
emmalynch: Does this make more sense to add as part of this issue: https://github.com/walmartlabs/cookie-cutter/issues/90

I'm working on it either way, just curious if it's better in this PR or as part of the integration test one.

sklose: k, makes sense to me

PR #121, opened 2020-05-27
Fix breaking API change in Azure Queues

plameniv:
The API contract of Azure Queue's was changed to accept milliseconds instead of seconds as part of a bugfix in 1.2. This PR reverts the breaking API change and restores semver compatibility.

### DIFF
--- a/packages/azure/src/streaming/index.ts
+++ b/packages/azure/src/streaming/index.ts
@@ -56,8 +56,8 @@ export interface IQueueSourceConfiguration {
     numOfMessages?: number;
 
     /**
-     * Required if not peek only. Specifies the new visibility timeout value, in milliseconds,
-     * relative to server time. The new value must be larger than or equal to 0, and cannot be larger than 7 days (604800000 milliseconds).
+     * Required if not peek only. Specifies the new visibility timeout value, in seconds,
+     * relative to server time. The new value must be larger than or equal to 0, and cannot be larger than 7 days (604800 seconds).
      * The visibility timeout of a message can be set to a value later than the expiry time.
      */
     visibilityTimeout?: number;
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-core",
-    "version": "1.2.0",
+    "version": "1.2.1",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
--- a/packages/core/src/__test__/config.test.ts
+++ b/packages/core/src/__test__/config.test.ts
@@ -104,6 +104,96 @@ describe("Primitive Values", () => {
         }
     }
 
+    interface ITimes {
+        readonly millisecondsToDays?: number;
+        readonly hoursToSeconds?: number;
+        readonly minutesToHours?: number;
+        readonly millisecondsToMinutes?: number;
+        readonly hoursToMinutes?: number;
+        readonly secondsToSeconds?: number;
+    }
+
+    @config.section
+    class Times implements ITimes {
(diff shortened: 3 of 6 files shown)


packages/core/src/__test__/config.test.ts
```diff
@@ -223,6 +313,36 @@ describe("Primitive Values", () => {
         }
     });
 
+    it("converts timespan -> number with different target and source units", () => {
+        const tester = {
```
sklose, line 317: nit: i think there is a cleaner way to write this test

```typescript
const input: ITimes = {
    millisecondsToDays: 2 * 24 * 60 * 60 * 1000
}

const expected: ITimes = {
    millisecondsToDays: 2
}

const actual = config.parse(Times, input);
expect(actual).toBe(expected);
```

packages/core/src/config.ts
```diff
@@ -234,34 +234,51 @@ export const converters = {
...
-            let millis = 0;
+            if (target < TimeSpanTargetUnit.Milliseconds || target > TimeSpanTargetUnit.Days) {
+                throw new Error(`unknown target unit '${target}'`);
+            }
+
+            if (source < TimeSpanTargetUnit.Milliseconds || source > TimeSpanTargetUnit.Days) {
+                throw new Error(`unknown target unit '${source}'`);
+            }
+            //                         x  ms    s   m   h   d
+            const conversionFactors = [0, 1, 1000, 60, 60, 24];
```
sklose, line 254: doesn't the `ms` library have all of these conversions implemented already? wondering if we can just leverage that instead of reimplementing it.

plameniv: I think `ms` only offers milliseconds to string and the reverse: https://github.com/zeit/ms#features

Other conversions would still need to be done by us. Maybe I can look for another library

plameniv: One example I found: https://github.com/ben-ng/convert-units

sklose: nvm, if `ms` doesn't do it then it's not worth pulling in yet another dependency

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-core",
```
sklose, line 2: fyi: you also need to add the change to azure queue config to the PR

plameniv: yeah, noted that when I posted in chat

packages/core/src/__test__/config.test.ts
plameniv: Tried some variations of the above, but keep getting this: `unexpected type, are you missing the '@section' decorator?`.

When I tried this: `expect({ ...actual }).toMatchObject({ ...expected });`, `{ ...actual}` comes out empty. This is the same decorator issue we have an open ticket about.

plameniv: Will make the suggested change but with explicit member comparisons: `expect(actual.millisecondsToDays).toBe(expected.millisecondsToDays);`

packages/azure/src/streaming/index.ts
```diff
@@ -56,8 +56,8 @@ export interface IQueueSourceConfiguration {
     numOfMessages?: number;
 
     /**
-     * Required if not peek only. Specifies the new visibility timeout value, in milliseconds,
```
sklose, line 59: when did this become milliseconds here in the API? thought this was seconds before 😕

sklose: found it, was part of @butorine 's change ... should be good to revert this back to seconds then

plameniv: Yes, now that azure's config is taking in seconds and not milliseconds

PR #126, opened 2020-06-02
Add RedisStreamSink & RedisStreamSource

dmmulroy:
# Introduction
This PR adds both an input sink and an output source to the redis package.  In addition to adding `RedisStreamSource` and `RedisStreamSink` classes, changes were made to `RedisClient` and `RedisProxy` to support stream commands.  Also, two new configuration options/types/interfaces for redis will be exported from the package; one for the Sink and one for the Source.  Changes to config will be detailed below.

## RedisStreamSink
The `RedisStreamSink` is a fairly straightforward `cookie-cutter` sink.  It can be created via a factory function, `redisStreamSink`, from the redis package.  The factory function takes a newly export config type named `IRedisInputStreamOptions`.  This is simply a union type between the default `IRedisOptions` type and new options for the sink:

```typescript
export type IRedisOutputStreamOptions = IRedisOptions & {
    readonly writeStream: string;
};
```

### RedisStreamSink Configuration Changes
- `writeStream` is the name of the stream that you want your client/application to write to

## RedisStreamSource
The `RedisStreamSource` is a bit more complicated than its counterpart.  For the initial release of redis stream support, the `RedisStreamSource` will by default be required to read from a consumer group rather than directly from the stream (you can find redis stream resources linked below). Generally speaking, `RedisStreamSource` works as follows in its `*start()` function:

 - On initial start, check the consumer group for messages that are pending for the current `consumerId`
 - On each loop/iteration do the following:
 - Check for any messages on the consumer groups PEL (Pending Entries List) that have exceeded the `idleTimout` config option.
 - Check for any new messages that have arrived into the consumer group since the last iteration
 - Process, `yield`, and `xack`  all of the above messages and move to the next iteration

 Like the sink, a factory function, `redisStreamSource`, is exported from the redis package.  The configuration options are bit more complex than sinks and are below:

```typescript
export type IRedisInputStreamOptions = IRedisOptions & {
    readonly readStream: string;
    readonly consumerGroup: string;
    readonly consumerId?: string;
    readonly consumerGroupStartId?: string;
    readonly blockTimeout?: number;
    readonly idleTimeout?: number;
    readonly batchSize?: number;
};
```

### RedisStreamSource Configuration Changes
- `readStream` is the name of the stream that you want your client/application to read from
- `consumerGroup` is the name of the consumer group that your client should join/be apart of
- `consumerId` is optional and is the name of your consumer for retrieving messages from the stream.  On initial startup the PEL for this client ID is checked/drained/processed. If this options is not supplied it will be generated via `short-id`
- `consumerGroupStartId` is optional, and in the case where your application/client is creating a new consumer group is the id used as a starting point in the stream. By default, `$` will be used.
- `blockTimeout` is optional and is the amount of time we block/wait on the `xreadgroup` command.  By default it will be 100ms.
- `idleTimeout` is optional and is the amount of time a message can sit idle on a PEL(read via `xreadgroup` but not yet acknowledged as processed via `xack`)  before having another consumer `xclaim` and process it. By default it will be 30s.
- `batchSize` is optional and is the number of messages your client/application will attempt to read via `xreadgroup` and `xclaim`.  By default it will be 10.

## Other changes
- `base64Encode` has been added to to `IRedisOptions`.  We were previously base64 encoding protos before we stored them in redis due to previous limitations of the `node-redis` library.  This has since changed, by default `base64Encode` will be true to preserve backwards compatibility.
- `jest.setTimeout` was adjusted to 10 seconds to prevent timeouts during the integration test suite
- Due to `@types/redis` being incomplete, we have to patch the types on our client in `RedisProxy`.

```typescript
interface IRedisCommandPatches {
    xadd: (key: string, id: string, ...args: (string | Buffer)[] | Callback<string>[]) => boolean;
    xreadgroup: (args: string[], cb: Callback<RawReadGroupResult>) => boolean;
    xgroup: (args: string[], cb: Callback<"OK">) => boolean;
    xack: (args: string[], cb: Callback<number>) => boolean;
    xpending: (args: string[], cb: Callback<RawPELResult>) => boolean;
    xclaim: (args: string[], cb: Callback<RawXClaimResult>) => boolean;
    set: (key: string, value: string | Buffer, cb: Callback<"OK">) => boolean;
}

export type RedisClientWithStreamOperations = RedisClient & IRedisCommandPatches;
```

## Test Coverage
For the initial release test coverage was focused on e2e integration tests and cover the following test cases:

- `RedisStreamSink` successfully adds a value to a redis stream through the output sink
- `RedisStreamSource` successfully creates a new consumer group
- `RedisStreamSource` successfully processes messages from the stream
- `RedisStreamSource` successfully processes expired idle messages from the consumer group

## Resources
- [redis streams introduction](https://redis.io/topics/streams-intro)
- [redis streams commands](https://redis.io/commands/xack)


packages/redis/src/RedisStreamSource.ts
```diff
@@ -0,0 +1,240 @@
...
+import { RedisOpenTracingTagKeys, RedisClient } from "./RedisClient";
+
+export class RedisStreamSource implements IInputSource, IRequireInitialization, IDisposable {
+    private done: boolean = false;
+    private client: Lifecycle<IRedisClient>;
+    private tracer: Tracer;
+    private spanOperationName: string = "Redis Input Source Client Call";
+
+    constructor(private readonly config: IRedisInputStreamOptions) {
+        this.tracer = DefaultComponentContext.tracer;
```
treyhoover, line 24: This could potentially be initialized on line 20

packages/redis/src/RedisStreamSource.ts
```diff
@@ -0,0 +1,240 @@
...
+
+                const messageRef = new MessageRef(
+                    { streamId: message.streamId },
+                    message,
+                    span.context()
+                );
+
+                messageRef.once("released", async (_, error) => {
+                    try {
+                        if (!error) {
```
treyhoover, line 62: Could just throw that error here instead of just checking for it, as the error handling is duplicated in the else and catch blocks

packages/redis/src/RedisStreamSource.ts
```diff
@@ -0,0 +1,240 @@
...
+
+                yield messageRef;
+            }
+
+            // Clear messages for next iteration
+            messages = [];
+        }
+    }
+
+    stop(): Promise<void> {
```
treyhoover, line 87: This one's not async, so return signature should probably just be `void`

packages/redis/package.json
```diff
@@ -13,19 +13,20 @@
...
-        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta",
-        "@types/redis": "2.8.16"
+        "@types/redis": "2.8.16",
+        "@walmartlabs/cookie-cutter-core": "^1.3.0-beta"
     },
     "scripts": {
         "build": "tsc",
         "test": "jest --config=../../jest.unit.config.js --rootDir=.",
-        "integrate": "docker-compose up -d && jest --config=../../jest.integration.config.js --rootDir=.; R=$?; docker-compose down; bash -c \"exit $R\"",
+        "integrate": "docker-compose up -d && jest --config=../../jest.integration.config.js --rootDir=. --detectOpenHandles; R=$?; docker-compose down; bash -c \"exit $R\"",
```
plameniv, line 29: Do you need the `--detectOpenHandles`?

dmmulroy: Ah nope, was using that to debug.

packages/redis/src/RedisClient.ts
```diff
@@ -153,4 +222,319 @@ export class RedisClient implements IRequireInitialization, IDisposable {
...
+                streamName,
+                consumerGroup,
+                result: RedisMetricResults.Success,
+            });
+            return response;
+        } catch (err) {
+            const alreadyExistsErrorMessage = "BUSYGROUP Consumer Group name already exists";
+            if (supressAlreadyExistsError && err.message === alreadyExistsErrorMessage) {
+                // TODO should this be considered an error or success?
+                this.metrics.increment(RedisMetrics.XGroupAlreadyExists, {
```
sklose, line 309: regarding the TODO ... do we even need a separate metric for this? can't think of any use case where somebody would be interested in this number ... cc @chrnola can you think of a use case?

packages/redis/src/RedisStreamSink.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    public async initialize(context: IComponentContext): Promise<void> {
+        this.client = makeLifecycle(redisClient(this.config));
+        await this.client.initialize(context);
+    }
+
+    public async dispose(): Promise<void> {
+        await this.client.dispose();
```
sklose, line 55: add a null check here in case initialize already failed

packages/redis/src/RedisStreamSource.ts
```diff
@@ -0,0 +1,234 @@
...
+
+                this.spanLogAndSetTags(
+                    span,
+                    this.config.db,
+                    this.config.readStream,
+                    this.config.consumerGroup
+                );
+
+                const messageRef = new MessageRef(
+                    { streamId: message.streamId },
```
sklose, line 50: i think you are still missing the remaining metadata here ... number of deliveries, timestamp, etc ..

dmmulroy: That data isn't available for Messages that aren't on the PEL and read via `XPENDING`. To implement that I'd have to reiterate over the message ids returned from xpending and both xreadgroups.

packages/redis/src/__test__/redisClient.integration.test.ts
```diff
@@ -12,47 +12,317 @@ import {
...
+            .input()
+            .add(redisStreamSource(cfg))
+            .done()
+            .dispatch({
+                onFoo: async (msg: Foo, ctx: IDispatchContext) => {
+                    ctx.publish(Bar, new Bar(`output for ${msg.text}`));
+                },
+            })
+            .output()
+            .published(new NullOutputSink())
```
sklose, line 304: nit: NullOutputSink is the default, you don't have to explicitly add it ... however there is a helper function `runIntegrationTest` that will collect all the msgs you publish from a handler and then you can assert that you got the correct data in the handler as well

packages/redis/src/RedisStreamSource.ts
sklose: k, i think you got to do that then ... @chrnola def needs the number of deliveries to decide to skip over a msg. Also when you add all the metadata do it with an enum that is exported from the package, so ppl can easily access it.

```
export enum StreamMetadata {
    StreamId = "streamId",
    NumberOfDeliveries = "num_of_deliveries"
}
```

and then in the handler ppl can do

```
function onSomeMessageType(msg: ISomeMsgType, ctx: IDispatchContext) {
    if (ctx.metadata<number>(StreamMetadata.NumberOfDeliveries)) > 10) {
        return;
    }
}
```

chrnola: I admittedly haven't thought through this fully yet, but, I think we can live without it for now. In practice we run our services such that all handler errors are fatal and I'd be comfortable with manually intervening if a consumer gets stuck. It's what we're used to doing with Kafka anyway.

packages/redis/src/RedisClient.ts
chrnola: I think it's worth tracking, but not with its own time series. I would vote for incrementing the `RedisMetrics.XGroupCreate` counter here but swapping the `reason` label for a new `RedisMetricResults.AlreadyExists` instead, something like:
```suggestion
                this.metrics.increment(RedisMetrics.XGroupCreate, {
                    ...
                    result: RedisMetricResults.AlreadyExists,
                });
```

packages/redis/src/RedisStreamSource.ts
sklose: K, we can skip adding the additional metadata then for now, but let's add the enum though so it's consistent with how other modules handle metadata.

packages/redis/src/__test__/redisClient.integration.test.ts
dmmulroy: Yeah I looked into using that, but felt that just asserting against the processed messages wasn't what I was looking for and was more a test that cookie-cutter itself was working rather than the pending message functionality.

PR #136, opened 2020-06-18
Add AMQP Sink + Source

plameniv:


packages/amqp/src/AmqpSink.ts
```diff
@@ -0,0 +1,69 @@
...
+            hostname: this.config.host,
+            port: this.config.port,
+        };
+        this.conn = await amqp.connect(options);
+        this.channel = await this.conn.createChannel();
+    }
+
+    public async sink(output: IterableIterator<IPublishedMessage>): Promise<void> {
+        const queueName = this.config.queueName;
+        const ok = await this.channel.assertQueue(queueName, { durable: true });
```
sklose, line 45: i think you can move this line to initialize ... should we make the `durable` flag configurable?

packages/amqp/src/AmqpSink.ts
```diff
@@ -0,0 +1,69 @@
...
+        this.logger.info("assertQueue", ok);
+
+        for (const msg of output) {
+            const payload = Buffer.from(this.config.encoder.encode(msg.message));
+            this.channel.sendToQueue(queueName, Buffer.from(payload), {
```
sklose, line 50: don't think you need `Buffer.from` twice

packages/amqp/src/AmqpSource.ts
```diff
@@ -0,0 +1,83 @@
...
+    public async initialize(context: IComponentContext): Promise<void> {
+        this.logger = context.logger;
+        const options: amqp.Options.Connect = {
+            protocol: "amqp",
+            hostname: this.config.host,
+            port: this.config.port,
+        };
+        this.conn = await amqp.connect(options);
+        this.channel = await this.conn.createChannel();
+        await this.channel.prefetch(1); // wait until message is acked before getting new one
```
sklose, line 44: i don't think we want to enable this ... this effectively turns the application into Serial mode no matter what ... it won't be able to pull any new messages until the current one is completely done (handler + sink)

packages/amqp/src/AmqpSource.ts
```diff
@@ -0,0 +1,83 @@
...
+        const ok = await this.channel.assertQueue(queueName, { durable: true });
+        this.logger.info("assertQueue", ok);
+
+        const pipe = this.pipe;
+        const ch = this.channel;
+        const encoder = this.config.encoder;
+        async function getMsg(msg: amqp.ConsumeMessage) {
+            if (msg !== null) {
+                const type = msg.properties.type;
+                const metadata: IMetadata = {};
```
sklose, line 59: we probably need to add some metadata here like "number of times msg was dequeued", etc.

packages/amqp/src/AmqpSource.ts
```diff
@@ -0,0 +1,83 @@
...
+                const codedMessage: IMessage = new EncodedMessage(encoder, type, msg.content);
+                const msgRef = new MessageRef(metadata, codedMessage, new SpanContext());
+                await pipe.enqueue(msgRef);
+                msgRef.once("released", async (_, err) => {
+                    if (!err) {
+                        ch.ack(msg);
```
sklose, line 65: is there a visibility window timeout for un-acked messages in rabbit? is that configurable?

packages/amqp/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface IAmqpConfiguration {
+    readonly host: string;
+    readonly port?: number;
+    readonly queueName: string;
+    readonly encoder: IMessageEncoder;
+}
+
+export function amqpSource(configuration: IAmqpConfiguration): IInputSource {
+    configuration = config.parse(AmqpConfiguration, configuration);
```
sklose, line 27: do you need to inject the default port here into the config since it's optional?

packages/amqp/src/AmqpSink.ts
plameniv: I think it should be fine to move it.

`durable: false` will make the queue and its messages disappear on broker restarts/crashes. I can see it being useful for testing purposes at least.

`persistent ` does the same for messages, where I think they disappear if the queue disappears and if they are marked as `persistent: false`.

I think it should be fine to add `durable` as optional parameter.

packages/amqp/src/AmqpSource.ts
plameniv: There is an `expiration` property which can be passed in.

packages/amqp/src/AmqpSource.ts
plameniv: The issue with removing this, is that one consumer will then always get the odd messages, for example.
Then if those are more work, the other worker will sit idle.

I think it makes sense to remove it but maybe we can find a solution that distributes messages to the non-busy consumers.

sklose: this should balance itself out as the consumers are running ... you control the prefetch count by setting the input queue size for CC and if one instance happens to handle more expensive messages then it will eventually produce back pressure and those messages will be processed by the other instance as well

packages/amqp/src/AmqpSource.ts
sklose: k, we should make that configurable as well

packages/amqp/src/index.ts
plameniv: I don't need to inject it, but I have added it since I refactored the config interface

packages/amqp/src/AmqpSource.ts
plameniv: Looks like there isn't a straightforward `dequeued` count metadata. There is something related to dead lettering but it only triggers on some more specific conditions like negative ack (which we are not doing), ttl expiration and queue limit reached.

Not sure if any of the available info is usable here.

```
export interface MessageProperties {
    contentType: any | undefined;
    contentEncoding: any | undefined;
    headers: MessagePropertyHeaders;
    deliveryMode: any | undefined;
    priority: any | undefined;
    correlationId: any | undefined;
    replyTo: any | undefined;
    expiration: any | undefined;
    messageId: any | undefined;
    timestamp: any | undefined;
    type: any | undefined;
    userId: any | undefined;
    appId: any | undefined;
    clusterId: any | undefined;
}
```

MessagePropertyHeaders has the dead-lettering related key/value pairs.

packages/amqp/src/AmqpSource.ts
```diff
@@ -0,0 +1,109 @@
...
+                    } catch (e) {
+                        failSpan(span, e);
+                        throw e;
+                    } finally {
+                        span.finish();
+                    }
+                });
+            }
+        }
+        await this.channel.consume(this.config.queue.queueName, getMsg, { noAck: false });
```
sklose, line 87: what are the semantics of noAck? does it mean the channel should not auto ack when the callback completes? in that case this would need to be true

packages/amqp/src/__test__/amqp.integration.test.ts
```diff
@@ -0,0 +1,109 @@
...
+                onMessageClass(msg: MessageClass, ctx: IDispatchContext<any>) {
+                    ctx.publish(MessageClass, msg);
+                },
+            })
+            .output()
+            .published(new CapturingOutputSink(outputOne))
+            .done()
+            .run(ErrorHandlingMode.LogAndContinue, ParallelismMode.Serial);
+
+        await sleep(200);
```
sklose, line 104: might wanna make this a `while outputOne.len() != numMessages` ... otherwise 200ms wait can make this test flaky

packages/amqp/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface IAmqpConfiguration {
+    readonly server?: IAmqpServerConfig;
+    readonly queue: IAmqpQueueConfig;
+    readonly message?: IAmqpMessageConfig;
+    readonly encoder: IMessageEncoder;
+}
+
+export function amqpSource(configuration: IAmqpConfiguration): IInputSource {
+    configuration = config.parse(AmqpConfiguration, configuration, {
+        server: { host: "localhost", port: 5672 }, // TODO: why does port not get passed to the output of parse
```
sklose, line 47: dont think we should default to localhost, host is kinda required to be passed in

packages/amqp/src/AmqpSource.ts
plameniv: `noAck: false` disables auto ack and enables manual ack.

Also some relevant info from RabbitMQ:

`In automatic acknowledgement mode, a message is considered to be successfully delivered immediately after it is sent.`
`Therefore, automatic message acknowledgement should be considered unsafe and not suitable for all workloads`

packages/amqp/src/index.ts
plameniv: I can't omit this since it's a required parameter. It's also not affected by Partial<T> because it's not top-level property.

We would need to have `RecursivePartial<T>` implemented if we want to fix this.

I can remove the entire `server` component from the default, however. That should remove the default host and port (especially since port isn't even getting through anyway)

PR #140, opened 2020-06-22
fix MsSqlSink throws wrong error

sklose:


packages/mssql/src/MssqlSink.ts
```diff
@@ -243,7 +243,14 @@ export class MssqlSink
                 database: this.config.database,
                 result: MssqlMetricResults.Error,
             });
-            await tx.rollback();
+            try {
+                await tx.rollback();
```
cross311, line 247: Should we do something like in example: https://github.com/tediousjs/node-mssql#transaction?  So that w know we should not rollback due to it already happening?

plameniv:
Do we want to completely ignore rollback errors?

packages/mssql/src/MssqlSink.ts
```diff
@@ -243,7 +243,14 @@ export class MssqlSink
...
+            } catch (e) {
```
talam, line 248: does this `e` overwrite the one above it?

cross311:
@sklose Looking at the event handling for rollback in docs, this might be the best method, but I at least wanted to put a comment about the docs talking about an rollback event begin raised that we could look at.

packages/mssql/src/MssqlSink.ts
sklose: don't think so, it's a different scope

packages/mssql/src/MssqlSink.ts
sklose: we could, though I can't think of a reason why we'd ever want to bubble up the rollback error if there was a previous error that is the reason why we are rolling back in the first place

plameniv: Do we have any actions we can take when a transaction was unsuccessfully rolled back?

If we hide the success/fail of a rollback would that not hide relevant information from the developer and result in some unknown state?

PR #144, opened 2020-06-26
Add metrics to AMQP

plameniv:


packages/amqp/src/AmqpSource.ts
```diff
@@ -97,13 +137,37 @@ export class AmqpSource implements IInputSource, IRequireInitialization, IDispos
...
     public async dispose(): Promise<void> {
         if (this.conn) {
             await this.conn.close(); // also closes channel
         }
     }
 
     public async stop(): Promise<void> {
+        this.running = false;
+        if (this.loop) {
+            clearTimeout(this.loop);
```
sklose, line 169: i think this has a race condition with the `loopAmqpQueueUnassignedMessageCount` where it reinstantiates the timer

packages/amqp/src/AmqpSource.ts
```diff
@@ -49,41 +66,64 @@ export class AmqpSource implements IInputSource, IRequireInitialization, IDispos
...
-        async function getMsg(msg: amqp.ConsumeMessage) {
-            const span = tracer.startSpan("Consuming Message For AMQP", {
+        const host = this.config.server.host;
+        const getMsg = async (msg: amqp.ConsumeMessage) => {
+            const event_type = msg.properties.type;
+            const span = this.tracer.startSpan("Consuming Message For AMQP", {
                 childOf: undefined,
             });
-            spanTags(span, "getMsg", queueName);
+            this.spanLogAndSetTags(span, "getMsg", queueName);
```
sklose, line 84: can you rename `getMsg` to something else? it's a bit of an odd name for a callback ... also i don't think we need to add it to the trace as a Fn name ... feels too much of an impl detail to explicitly name a callback in the trace

plameniv: The function's name is `onMessage` in the consume's parameter list, so I can rename it to that.

Will remove from the trace.

packages/amqp/src/AmqpSource.ts
```diff
@@ -49,61 +65,109 @@ export class AmqpSource implements IInputSource, IRequireInitialization, IDispos
...
                     } catch (e) {
                         failSpan(span, e);
                         throw e;
                     } finally {
                         span.finish();
+                        this.metrics.increment(AmqpMetrics.MsgProcessed, {
+                            host,
+                            queueName,
+                            event_type,
+                            result,
```
chrnola, line 120: nit: it seems to me like `result` has more to do with whether an `ack` succeeds or fails, as opposed to the processing of the message as the name of the metric would suggest.

sklose: the span is for the entire flow of receiving to routing through CC and then acking it

chrnola: Fair enough, but what would happen if the CC handler does not throw, but `this.channel.ack(msg)` does? The message would (eventually) be delivered again?

I'm just wondering if there's value in counting ack success/failures separately. :thinking:

plameniv: The message gets delivered again, yes.

plameniv: I'm not sure what the frequency of `ack` failures is going to be.

Another thing is we can `negatively ack` as well. I haven't done any significant testing on the effects of `nack` vs not `ack`.

Will do some testing.

Issue #145, opened 2020-06-27
lz4 package is broken


k-gupta:
`lz4 0.6.4`  package is missing some parts - https://github.com/pierrec/node-lz4/issues/99. This causes errors when installing `cookie-cutter-kafka` package due to its dependency to `kafkajs-lz`.

PR #146, opened 2020-06-30
Update BlobClient to use latest azure storage library

emmalynch:
* Replace azure-storage with @azure/blob-storage library
* Remove timeout options, as timeouts are handled by the library
* Adds resolution for lz4 to pin to the previous version - fixes this issue: https://github.com/walmartlabs/cookie-cutter/issues/145


packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -1,24 +1,33 @@
 /*
 Copyright (c) Walmart Inc.
-
```
plameniv, line 3: Add this line back in, so we are consistent across all files. Need to change this in a couple of places.

packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -32,70 +41,137 @@ describe("BlobClient", () => {
...
+        };
+        const context = new SpanContext();
+        const span: Span = new NullTracerBuilder()
+            .create()
+            .startSpan("unit-test", { childOf: context });
         it("performs a successful read", async () => {
             const blobClient = new BlobClient(config);
             await expect(blobClient.read(span.context(), "BlobID")).resolves.toBe(text);
         });
-
```
plameniv, line 91: Add an empty line before each unit test.

packages/azure/src/utils/BlobClient.ts
```diff
@@ -66,16 +58,18 @@ export class BlobClient implements IRequireInitialization {
     public async createContainerIfNotExists(context?: SpanContext) {
         const span = this.tracer.startSpan(this.spanOperationName, { childOf: context });
         this.spanLogAndSetTags(span, this.createContainerIfNotExists.name);
-        return new Promise<BlobService.ContainerResult>((resolve, reject) => {
-            this.blobService.createContainerIfNotExists(this.containerName, (error, result) => {
-                if (error) {
+        return new Promise<ContainerCreateResponse>((resolve, reject) => {
+            this.blobService
+                .createContainer(this.containerName)
```
sklose, line 63: can we simplify all the `new Promise( .. .then().catch() )` chains into simple `try { await x } catch { } finally { }` blocks?

packages/azure/src/utils/CosmosClient.ts
```diff
@@ -23,7 +23,7 @@ import {
     OpenTracingTagKeys,
 } from "@walmartlabs/cookie-cutter-core";
 import * as fs from "fs";
-import { Agent, AgentOptions } from "https";
+import { Agent, AgentOptions } from "http";
```
sklose, line 26: changing from https to http is on purpose?

packages/azure/src/utils/BlobClient.ts
emmalynch: Will do. For my own knowledge, is this just a stylistic preference, or is there some other advantage to using try/catch in this scenario?

sklose: yes, it's purely stylistic ... we use the try/await/catch approach everywhere else and I'd prefer to keep the code base consistent

packages/azure/src/utils/CosmosClient.ts
emmalynch: Fixed!

packages/azure/src/utils/BlobClient.ts
emmalynch: Makes sense, I've switch to try/catch everywhere now

packages/azure/src/__test__/utils/BlobClient.test.ts
emmalynch: done!

packages/azure/src/__test__/utils/BlobClient.test.ts
emmalynch: done

packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -5,20 +5,30 @@ This source code is licensed under the Apache 2.0 license found in the
...
 import { NullTracerBuilder } from "@walmartlabs/cookie-cutter-core";
-import { createBlobService } from "azure-storage";
 import { Span, SpanContext } from "opentracing";
 import { IBlobStorageConfiguration } from "../..";
 import { BlobClient } from "../../utils";
+import { BlobServiceClient } from "@azure/storage-blob";
+import { streamToString } from "../../utils/helpers";
+
+const MockBlobServiceClient: jest.Mock = BlobServiceClient as any;
```
sklose, line 15: for my understanding: how does the mocking actually work here?

```
const MockStreamToString: jest.Mock = streamToString as any;

jest.mock("../../utils/helpers", () => {
    return {
        streamToString: jest.fn().mockImplementation(() => jest.fn()),
    };
});
```

first it assigns the original `streamToString` function with a forced-cast to `MockStreamToString` ... does that mean `MockStreamToString` is the original function? Mocking the module happens only afterwards ... 😕

packages/azure/src/utils/BlobClient.ts
```diff
@@ -99,113 +90,104 @@ export class BlobClient implements IRequireInitialization {
...
-                        resolve();
-                    }
-                }
+        const containerClient = this.blobService.getContainerClient(this.containerName);
+        const blobClient = containerClient.getBlockBlobClient(blobId);
+
+        try {
+            const result = await blobClient.upload(text, Buffer.byteLength(text));
+
+            if (result?._response?.status) {
```
sklose, line 102: `_response` looks a little suspicious to me ... does the _ indicate that we are using an internal field here? is there an official API to get the status code?

packages/azure/src/__test__/utils/BlobClient.test.ts
emmalynch: The module mocking will be hoisted automatically, but if it's clearer to read, I can move it to the top of the file. I've assigned the mock to a const so that I can mock the implementation later.

packages/azure/src/utils/BlobClient.ts
emmalynch: The upload method returns the following object: https://docs.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blockblobuploadresponse?view=azure-node-latest

Looking at the actual code, there's no private modifier on the field: https://github.com/Azure/azure-sdk-for-js/blob/410dbb845aed4fde2a4c1c474afeb5d0f0a467c6/sdk/storage/storage-blob/src/generated/src/models/index.ts#L8151

None of the samples I looked at in the repo accessed the status from the response, only from the errors, and I couldn't find another way of accessing it from reading the docs.

plameniv:
Changes were incorporated in #204

PR #148, opened 2020-07-03
Update QueueClient to use latest azure storage library

emmalynch:
Update QueueClient to use latest azure-storage library

### DIFF
--- a/packages/azure/src/utils/CosmosClient.ts
+++ b/packages/azure/src/utils/CosmosClient.ts
@@ -25,6 +25,7 @@ import {
 import * as fs from "fs";
 import { Agent, AgentOptions } from "https";
 import { FORMAT_HTTP_HEADERS, Span, SpanContext, Tags, Tracer } from "opentracing";
+import * as http from "http";
 import * as path from "path";
 import * as tunnel from "tunnel";
 import * as url from "url";
@@ -48,6 +49,8 @@ export interface ICosmosWriteClient {
     ): Promise<void>;
 }
 
+type AgentWithOptions = http.Agent & { options?: AgentOptions };
+
 enum CosmosMetrics {
     RUs = "cookie_cutter.cosmos_client.request_units",
     Sproc = "cookie_cutter.cosmos_client.execute_sproc",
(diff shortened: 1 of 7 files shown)


packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -1,24 +1,33 @@
 /*
 Copyright (c) Walmart Inc.
-
```
plameniv, line 3: Add this line back in, so we are consistent across all files. Need to change this in a couple of places.

packages/azure/src/utils/BlobClient.ts
```diff
@@ -102,110 +96,109 @@ export class BlobClient implements IRequireInitialization {
...
-                            BlobMetrics.Exists,
-                            this.generateMetricTags(BlobMetricResults.Error, statusCode)
-                        );
-                        failSpan(span, err);
-                        span.finish();
-                        reject(err);
-                    } else {
+        const containers = this.blobService.listContainers();
+
+        return new Promise<boolean>(async (resolve, reject) => {
```
sklose, line 173: you can remove the promise here and just make the function async since you are using await here anyways

plameniv:
Looks like this PR contains all the changes from #146. Do we need both PRs?

packages/azure/src/utils/CosmosClient.ts
```diff
@@ -23,7 +23,7 @@ import {
     OpenTracingTagKeys,
 } from "@walmartlabs/cookie-cutter-core";
 import * as fs from "fs";
-import { Agent, AgentOptions } from "https";
+import { Agent, AgentOptions } from "http";
```
plameniv, line 26: This looks like an issue I ran into where I had `node_modules`, `yarn.lock` and `packages.json` in directly in my user directory. This was causing some dependency issues around some property that was not matching between the `https` and `http` Agent classes.

If this is the same issue, I solved it by removing those files.

emmalynch:
@plameniv you're right that this also contains the changes from the other PR, going to stick this one back to draft mode until the other one is finished up

plameniv:
Changes were incorporated in #204

PR #155, opened 2020-08-05
cookie-cutter-redis: add support for multiple streams

mag415:


packages/redis/src/RedisStreamSource.ts
```diff
@@ -151,35 +153,36 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
+                    this.config.consumerGroup,
+                    this.config.batchSize
+                );
 
-            // if there are no pending messages return early w/ an empty array
-            if (pendingMessages.length < 1) {
-                return [];
-            }
+                // if there are no pending messages return early w/ an empty array
+                if (pendingMessages.length < 1) {
```
prachi30, line 169: nit: `if (!pendingMessages.length)`

mag415: updated

packages/redis/src/RedisClient.ts
```diff
@@ -367,19 +367,19 @@ export class RedisClient implements IRedisClient, IRequireInitialization, IDispo
...
     public async xReadGroup(
         context: SpanContext,
-        streamName: string,
+        streamNames: string[],
         consumerGroup: string,
         consumerName: string,
         count: number,
         block: number,
-        id: string = ">"
+        ids?: string[]
```
sklose, line 375: it looks like `ids` isn't actually used anymore. couple of options (not sure what makes most sense here)

1. keep it as `id: string = ">"` and change `streamNames.map(() => ">")` to `streamNames.map(() => id)`
2. streamName + id seem to come in pairs, maybe combine that as a single argument `streams: { name: string; id?: string }[]`;

mag415: updated to use suggestion #2

PR #156, opened 2020-08-07
cookie-cutter-redis: add metrics

mag415:


packages/redis/src/RedisStreamSource.ts
```diff
@@ -35,6 +44,9 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
             const newMessages = await this.getNewMessages();
             messages.push(...newMessages);
 
+            this.metrics.gauge(RedisMetrics.IncomingBatchSize, messages.length, {
+                stream_names: this.config.readStreams,
```
sklose, line 48: don't think you can have array's in metrics ... labels need to be single values

mag415: ah right. I'll remove that tag

PR #164, opened 2020-08-17
fix pending list not fully drained on startup, added password config

sklose:
PR additionally contains refactoring to remove double tracing

### DIFF
--- a/packages/redis/src/RedisStreamSource.ts
+++ b/packages/redis/src/RedisStreamSource.ts
@@ -34,7 +34,7 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization, 
 
     public async *start(): AsyncIterableIterator<MessageRef> {
         // On initial start attempt to fetch any messages from this consumers PEL
-        // these are messages that were previously read with XGroupRead by
+        // these are messages that were previously read with XReadGroup by
         // this consumer group, but were not acked. This can happen on service restarts
         const streams = this.config.readStreams.map((name) => ({ name, id: "0" }));
         while (!this.done) {
@@ -63,7 +63,7 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization, 
 
                 this.metrics.gauge(RedisMetrics.IncomingBatchSize, messages.length, {});
                 for (const message of messages) {
-                    // when calling xReadGroup again only get messages after this one
+                    // when calling XReadGroup again only get messages after this one
                     streams.filter((s) => s.name === message.streamName)[0].id = message.streamId;
                     yield this.createMessageRef(message, span);
                 }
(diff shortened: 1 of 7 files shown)


packages/redis/src/RedisStreamSource.ts
```diff
@@ -26,102 +28,116 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
+                this.metrics.gauge(RedisMetrics.IncomingBatchSize, messages.length, {});
+                for (const message of messages) {
+                    // when calling xReadGroup again only get messages after this one
+                    streams.filter((s) => s.name === message.streamName)[0].id = message.streamId;
+                    yield this.createMessageRef(message, span);
+                }
+            } catch (e) {
+                failSpan(span, e);
+                this.logger.error("failed to read messages from Redis, retrying in 500ms", e);
+                await sleep(500);
```
chrnola, line 76: Trying to think through what would happen when the server was restarted (for volatile deployments) or the stream key was evicted or otherwise removed. I'm reasonably certain that we could get stuck retrying here forever. :thinking:

Note that `XREADGROUP` could fail because the consumer group no longer exists (volatile Redis deployment restarted/stream key evicted). At the moment we handle re-creating the group, but only in `initialize()`.

I'm thinking it may be safer to crash instead.

packages/redis/src/RedisStreamSource.ts
```diff
@@ -26,102 +28,116 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
+                span,
+                this.config.db,
+                this.config.readStreams,
+                this.config.consumerGroup
+            );
+
+            try {
+                let messages = [];
+
+                // Once in a while check the PEL list for _other_ consumer groups
```
chrnola, line 97: Small nitpick:
```suggestion
                // Once in a while check the PEL list for _other_ consumers
```

packages/redis/src/RedisStreamSource.ts
```diff
@@ -26,102 +26,113 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
     private spanOperationName: string = "Redis Input Source Client Call";
     private lastPendingMessagesCheck: Date | undefined;
 
     constructor(private readonly config: IRedisInputStreamOptions) {}
 
     public async *start(): AsyncIterableIterator<MessageRef> {
         // On initial start attempt to fetch any messages from this consumers PEL
-        let messages = await this.getPendingMessagesForConsumer();
-
+        // these are messages that were previously read with XGroupRead by
```
plameniv, line 37: nit: should this be `xReadGroup`

Issue #96, opened 2020-04-20
The config's parse function returns an object with no enumerable properties.


plameniv:
The config object returned by the [parse](https://github.com/walmartlabs/cookie-cutter/blob/d7726098213ce5966ba920a3fca7fb5890ba4b6c/packages/core/src/config.ts#L59) function has no enumerable properties.

The following operations do not work:
- `console.log(config)` prints the empty object;
- `const newConfig = { ...config }` creates an empty object;
- `Object.keys(config)` and `Object.getOwnPropertyNames(config)` don't return any values.

When the property is manually defined in the following manner, the converter functions stop working, i.e. `1m` is not converted to 60 (seconds):

```
Object.defineProperty(target, key, {
     enumerable: true,
     writable: true,
     value: source[key],
});
target.__assignedProperties.add(key);
```

plameniv:
https://github.com/walmartlabs/cookie-cutter/pull/69 addressed symptoms of this bug.

Once the bug is fixed, some of the changes in that PR may be reverted.

PR #171, opened 2020-08-21
cleanup redis stream implementation

sklose:


packages/redis/src/index.ts
```diff
@@ -133,31 +120,48 @@ export interface IRedisClient {
...
+        db: 0,
+        base64Encode: true,
+        payloadKey: "redis.stream.key",
+        typeNameKey: "redis.stream.type",
+        typeMapper: new ObjectNameMessageTypeMapper(),
+    });
     return new RedisStreamSink(configuration);
 }
 
 export function redisStreamSource(configuration: IRedisInputStreamOptions): IInputSource {
```
chrnola, line 151: It may be worth pointing out that we started calling the `RedisStreamSource` constructor directly as a workaround for #96.

I unfortunately don't recall what was special about the Redis module though, since we typically _do_ use the wrapper functions for other cookie-cutter sources... 🤔

sklose: the issues was that the sink used the `redisClient` function which ended up doing a double parse on the config. that should be fixed now and you can use the function instead

packages/redis/src/RedisClient.ts
```diff
@@ -234,34 +232,36 @@ export class RedisClient implements IRedisClient, IRequireInitialization, IDispo
...
+            });
 
-        this.spanLogAndSetTags(span, this.xAddObject.name, db, key, streamName);
+            const buf = Buffer.from(encodedBody);
+            const storableValue = this.config.base64Encode ? buf.toString("base64") : buf;
 
-        const typeName = this.getTypeName(type);
+            const args: (string | Buffer)[] = [streamName];
+            if (!isNullOrUndefined(maxStreamLength)) {
+                args.push(...["MAXLEN", "~", maxStreamLength.toString()]);
```
plameniv, line 258: Can this just be `args.push("MAXLEN", "~", maxStreamLength.toString());`?
Same with the next `args.push`

packages/redis/src/RedisStreamSource.ts
```diff
@@ -99,30 +104,35 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
                 // to reclaim orphaned messages. This can happen if a consumer dies
                 // or permanently leaves the consumer groupe (e.g. scaling down to less instances)
                 if (
-                    this.lastPendingMessagesCheck === undefined ||
-                    this.lastPendingMessagesCheck.getTime() + this.config.reclaimMessageInterval <=
-                        Date.now()
+                    this.config.reclaimMessageInterval &&
+                    didXReadGroup &&
+                    (this.lastPendingMessagesCheck === undefined ||
+                        this.lastPendingMessagesCheck.getTime() +
```
plameniv, line 110: Is this the proper indentation?

packages/redis/src/RedisClient.ts
sklose: yep, good catch

packages/redis/src/RedisStreamSource.ts
sklose: it's what prettier wants it to be ... if i format it any other way we'd get a linter error unless we disable lint for that block

packages/redis/src/RedisStreamSource.ts
```diff
@@ -213,48 +215,42 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
-            this.logger.debug(
-                "acking stream=" + streamName + " id=" + streamId + " group=" + consumerGroup
-            );
-            const count = await this.client.xAck(
-                span.context(),
-                streamName,
-                consumerGroup,
-                streamId
-            );
+            const count = await this.client.xAck(span.context(), stream, consumerId, messageId);
```
plameniv, line 225: why did the return type of `xack` change from `boolean` in `IRedisCommandPatches` to `number` here?

`count` should probably just be called `found`

packages/redis/src/RedisStreamSource.ts
```diff
@@ -213,48 +215,42 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
...
             failSpan(span, e);
             this.metrics.increment(RedisMetrics.MsgProcessed, {
-                stream_name: streamName,
-                consumer_group: consumerGroup,
+                stream_name: stream,
+                consumer_group: consumerId,
                 result: RedisMetricResult.Error,
             });
+
+            if (!err) {
```
plameniv, line 243: Is it possible for an error to be caught that comes from `this.client.xAck` and if so, why are we ignoring it?

plameniv: By ignoring I mean, we don't log it.

sklose: I just don't want to log if the error came from a message handler ... that is part of normal flow, only log if the error happened within here

packages/redis/src/RedisStreamSource.ts
sklose: you can technically `xack` multiple messages in one shot and it returns the number it successfully acked. `boolean` would probably work out since 0/1 in javascript are truthy/falsey, but it wouldn't be quite the right type

plameniv: I think I misread the redis `xack` function where the callback returns a number.

I assume we can't get more than 1 message `xack` with the current code?

sklose: right, we currently don't support batching when releasing messages

packages/redis/src/__test__/stream.integration.test.ts
```diff
@@ -0,0 +1,329 @@
...
+            const consumer = Application.create()
+                .input()
+                .add(
+                    redisStreamSource({
+                        consumerGroup: `roundtrip-test-group-${id}`,
+                        encoder: new JsonMessageEncoder(),
+                        host: "localhost",
+                        streams: [`roundtrip-test-1-${id}`, `roundtrip-test-2-${id}`],
+                        typeMapper: new ObjectNameMessageTypeMapper(),
+
```
plameniv, line 119: do we need the empty lines here and in the other consumers?

packages/redis/src/__test__/stream.integration.test.ts
```diff
@@ -0,0 +1,329 @@
...
+    beforeAll(() => {
+        jest.setTimeout(90000);
+    });
+
+    for (const [id, cfg] of RoundTripTestConfigurationPermutations) {
+        it(`produces and consumes messages - ${id}`, async () => {
+            const input: MessageRef[] = [];
+            for (let i = 0; i < 25; i++) {
+                input.push(
+                    ...[
```
plameniv, line 59: No need to `...` an array here.

packages/redis/src/__test__/stream.integration.test.ts
sklose: i moved it around a bit

PR #182, opened 2020-09-02
Fix backwards compatibility for ProtoMessageEncoder

plameniv:


packages/proto/src/__test__/ProtoMessageEncoder.test.ts
```diff
@@ -59,4 +60,16 @@ describe("ProtoMessageEncoder", () => {
         expect(base64Decoded).toMatchObject(encodedMsg);
         expect(decodedMsg).toMatchObject(msg);
     });
+
+    it("returns correct Uint8Array from ", async () => {
```
sklose, line 64: nit: you are missing a word after "from"

packages/proto/src/__test__/ProtoMessageEncoder.test.ts
```diff
@@ -59,4 +60,16 @@ describe("ProtoMessageEncoder", () => {
...
+        const root = await loadTestProto();
+        const encoder = new ProtoMessageEncoder(pbjsStaticModuleRegistry(root), false);
+        const encodedData = msg;
+        const bufferOfEncodedData = encoder.encode(encodedData);
+        // in cosmosdb 'val: <buffer>' becomes: 'val: { type: "Buffer", data: <buffer> }'
+        const cosmosEncodedData = { type: "Buffer", data: bufferOfEncodedData };
+        const buffer = encoder.fromJsonEmbedding(cosmosEncodedData);
```
sklose, line 71: wondering if this test is 100% valid ... 

```
const bufferOfEncodedData = encoder.encode(encodedData);
```
`bufferOfEncodedData` is of type Uint8Array here (the proper type)


```
        const cosmosEncodedData = { type: "Buffer", data: bufferOfEncodedData };
        const buffer = encoder.fromJsonEmbedding(cosmosEncodedData);
```
here you are passing in the properly typed `bufferOfEncodedData` for `data` and the `fromJsonEmbedding` returns `.data` and that way the types will def align perfectly.

Should we do a JSON.stringify/parse roundtrip in between to force the types to disappear to be on the safe side with this test?

plameniv: So, stringify and then parse `cosmosEncodedData`? 

I don't know how much it will improve the test. `fromJsonEmbedding` accepts type `any` and it's internals treat the input as `any`.

plameniv: Actually, `fromJsonEmbedding` treats it as if it has `.data` of type Buffer. Type information is not used.

sklose (approved):

PR #186, opened 2020-09-08
detect when kafkajs is stuck with stale broker metadata

sklose:
This PR detects stale metadata cached in kafkajs and forces the application terminate as it currently gets stuck in an infinite error/retry loop. This is only intended to be a workaround until the underlying root cause has been addressed.

Only planning to merge this if needed

### DIFF
--- a/packages/kafka/src/KafkaConsumer.ts
+++ b/packages/kafka/src/KafkaConsumer.ts
@@ -416,7 +416,10 @@ export class KafkaConsumer implements IRequireInitialization, IDisposable {
 
                     // this is a workaround for https://github.com/walmartlabs/cookie-cutter/issues/185
                     // until a fix for kafkajs is available / the root cause of the problem is confirmed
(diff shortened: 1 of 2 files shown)


packages/kafka/src/KafkaConsumer.ts
```diff
@@ -407,15 +409,36 @@ export class KafkaConsumer implements IRequireInitialization, IDisposable {
...
+                    if (
+                        e &&
+                        e.toString().contains("server is not the leader for that topic-partition")
+                    ) {
+                        this.brokerMetadataErrors++;
+                    }
+
+                    if (this.brokerMetadataErrors > 10) {
+                        this.logger.error("detected stale broker metadata in kafkajs");
+                        this.brokerMetadataErrors = 0;
```
plameniv, line 428: Is this needed if the application terminates after the thrown error?

plameniv:
Merging this in to see how well it addresses the issue since we still don't know what the underlying issue is.

plameniv (approved):

PR #192, opened 2020-09-18
Add auth for amqp source and sink

prachi30:


packages/amqp/src/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as amqp from "amqplib";
+export function getAmqpConnectionConfig(config: IAmqpServerConfig): amqp.Options.Connect {
+    const options: amqp.Options.Connect = {
+        protocol: "amqp",
+        hostname: config.host,
+        port: config.port,
+    };
+
+    // specify username and password only when provided in config to avoid overwriting the defaults
+    if (config.username) {
```
plameniv, line 11: Maybe we should check if both are set? Not sure what the behavior is if only the username is provided.

packages/amqp/src/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+        return {
+            ...options,
+            username: config.username,
+            password: config.password, 
+        };
+    }
+    return options;
+}
```
plameniv, line 19: lint: add empty line at end of file

plameniv:
I think for the change to be properly tested we’d need to also modify the docker image to set up a new username/password and use those in the integration test.

I don’t know if it can be done with docker-compose, but we can definitely make our own Dockerfile with appropriate setup commands.

Old Example here: https://stackoverflow.com/questions/30747469/how-to-add-initial-users-when-starting-a-rabbitmq-docker-container

packages/amqp/src/utils.ts
prachi30: actually, even username is not required based on the underlying lib implementation

plameniv (approved):

PR #203, opened 2020-10-28
Don't extract data field in QueueInputSource when encoder isEmbeddable

plameniv:


packages/azure/src/streaming/internal/QueueInputSource.ts
```diff
@@ -98,7 +98,12 @@ export class QueueInputSource implements IInputSource, IRequireInitialization {
                 };
                 const event_type = headers[EventSourcedMetadata.EventType];
                 let protoOrJsonPayload = payload;
-                if (payload.type && payload.type === "Buffer" && isArray(payload.data)) {
+                if (
```
sklose, line 101: do we need this if-statement at all? i feel like it would still work for all cases if we just removed it

plameniv: The decode function of non Embeddable encoders expects to receive an `Uint8Array`, which in the case of a Buffer is going to be in `payload.data`. So, if we remove the `if-statement` it will get the payload which would be of type `{ type: string; data: Uint8Array }`, I think.

sklose: k, I guess we can leave it to be on the safe side. I think practically it isn't used anymore. All our encoders we ship do implement that embedding interface now.

sklose (approved):

PR #204, opened 2020-11-02
Azure module integration tests

plameniv:


packages/azure/package.json
```diff
@@ -30,7 +31,7 @@
         "build": "tsc",
         "postbuild": "node postbuild.js",
         "test": "jest --config=../../jest.unit.config.js --rootDir=.",
-        "integrate": "docker-compose up -d && echo \"no tests yet\"; R=$?; docker-compose down; bash -c \"exit $R\"",
+        "integrate": "if [[ $RUNNING_IN_CI != 1 ]]; then sh setup_env_vars_locally.sh; cd vm; st=$(vagrant status); vagrant up; if [[ ! $st =~ \"running\" ]]; then sleep 30s; fi; cd ..; fi; jest --config=../../jest.integration.config.js --rootDir=.",
```
sklose, line 34: do you need a `vagrant down` here in the end for local? maybe worth extracting into a bashscript here ... getting a little long for a one liner

plameniv: For local running of `integrate` if I did `vagrant halt` then every time the integration command is ran you'd have to wait for the VM to boot and then wait a bit so you can run the tests themselves with no errors.

I assume locally the developer would want to iterate over changes and not have to wait for the spin up process every time. That's also why I have one of the `if` statements.

I think it's a good idea to move the command into a script. Will do that.

sklose: can you control it with a an argument? i believe if you run `yarn run integration --keep` then `--keep` should be passed to your script ... so default behavior would be to cleanup, and if you need to keep it up for iterating you can add the arg

plameniv: Will look into it.

sklose (approved):

PR #212, opened 2021-02-01
Updates to Azure BlobClient with more methods

chandanRIT:
Update Azure BlobClient with new methods to list blobs, delete blob, delete folder, write Large blob

### DIFF
--- a/packages/azure/src/__test__/utils/BlobClient.test.ts
+++ b/packages/azure/src/__test__/utils/BlobClient.test.ts
@@ -269,6 +269,7 @@ describe("BlobClient", () => {
             });
 
             const blobClient: BlobClient = new BlobClient(config);
+            /* tslint:disable-next-line:no-floating-promises */
             expect(
                 blobClient.listAllBlobs(blobFolderPrefix, null, span.context())
             ).resolves.toEqual([`${blobFolderPrefix}/a`, `${blobFolderPrefix}/b`]);
@@ -312,6 +313,7 @@ describe("BlobClient", () => {
                 );
 
             const blobClient: BlobClient = new BlobClient(config);
+            /* tslint:disable-next-line:no-floating-promises */
             expect(
                 blobClient.listAllBlobs(blobFolderPrefix, null, span.context())
             ).resolves.toEqual([
@@ -334,6 +336,7 @@ describe("BlobClient", () => {
             });
 
             const blobClient: BlobClient = new BlobClient(config);
+            /* tslint:disable-next-line:no-floating-promises */
             expect(
                 blobClient.listAllBlobs("BlobFolderPrefix", null, span.context())
             ).rejects.toEqual(error);
--- a/packages/azure/src/__test__/utils/BlobClient.test.ts
+++ b/packages/azure/src/__test__/utils/BlobClient.test.ts
@@ -12,8 +12,17 @@ jest.mock("azure-storage", () => {
     };
 });
 
+jest.mock("fs", () => {
+    return {
+        promises: {
+            writeFile: jest.fn(),
+            unlink: jest.fn(),
+        },
+    };
+});
+
 import { NullTracerBuilder } from "@walmartlabs/cookie-cutter-core";
-import { createBlobService } from "azure-storage";
+import { BlobService, createBlobService } from "azure-storage";
 import { Span, SpanContext } from "opentracing";
 import { IBlobStorageConfiguration } from "../..";
 import { BlobClient } from "../../utils";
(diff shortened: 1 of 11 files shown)


packages/azure/src/utils/BlobClient.ts
```diff
@@ -14,10 +14,13 @@ import {
     IRequireInitialization,
     OpenTracingTagKeys,
 } from "@walmartlabs/cookie-cutter-core";
-import { BlobService, createBlobService, ServiceResponse } from "azure-storage";
```
chandanRIT, line 17: add tests

plameniv:
Bump the version number in `package.json`

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
     configuration = config.parse(CosmosConfiguration, configuration);
     return new CosmosClient(configuration);
 }
+
+export interface IBlobClient {
+    initialize(context: IComponentContext);
```
sklose, line 56: `initialize` should not be on this interface, there is a dedicated interface defining it `IRequireInitialization`

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    createContainerIfNotExists(context?: SpanContext): Promise<BlobService.ContainerResult>;
+    write(context: SpanContext, text: Buffer | string, blobId: string): Promise<void>;
+    read(context: SpanContext, blobId: string): Promise<string>;
+    exists(context: SpanContext, blobId: string): Promise<boolean>;
+    deleteFolderIfExists(folderSubPath: string, context: SpanContext): Promise<boolean>;
+    listAllBlobs(
+        prefix: string,
+        continuationToken: common.ContinuationToken,
```
sklose, line 64: this is a leaky abstraction ... you are re-exporting a type from the azure storage library here ... we should not do that

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
     configuration = config.parse(CosmosConfiguration, configuration);
     return new CosmosClient(configuration);
 }
+
+export interface IBlobClient {
+    initialize(context: IComponentContext);
+    createContainerIfNotExists(context?: SpanContext): Promise<BlobService.ContainerResult>;
+    write(context: SpanContext, text: Buffer | string, blobId: string): Promise<void>;
+    read(context: SpanContext, blobId: string): Promise<string>;
```
sklose, line 59: maybe rename to `readAsText` to leave it open to future use cases like `readAsBuffer`

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    exists(context: SpanContext, blobId: string): Promise<boolean>;
+    deleteFolderIfExists(folderSubPath: string, context: SpanContext): Promise<boolean>;
+    listAllBlobs(
+        prefix: string,
+        continuationToken: common.ContinuationToken,
+        context: SpanContext
+    ): Promise<string[]>;
+    deleteBlobIfExists(blobSubPath: string, context: SpanContext): Promise<boolean>;
```
sklose, line 67: can you unify terminology here ... `blobSubPath`, `folderSubPath`, `blobId` ...

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    writeLargeObject(obj: any, blobId: string, context: SpanContext): Promise<void>;
+}
+
+export function createBlobClient(config: IBlobStorageConfiguration): IBlobClient {
+    return new BlobClient(config);
```
sklose, line 72: I believe this needs to run thru the config parser library like in other factory functions so all the value shortcuts work

packages/azure/src/index.ts
```diff
@@ -49,3 +51,23 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    read(context: SpanContext, blobId: string): Promise<string>;
+    exists(context: SpanContext, blobId: string): Promise<boolean>;
+    deleteFolderIfExists(folderSubPath: string, context: SpanContext): Promise<boolean>;
+    listAllBlobs(
+        prefix: string,
+        continuationToken: common.ContinuationToken,
+        context: SpanContext
+    ): Promise<string[]>;
+    deleteBlobIfExists(blobSubPath: string, context: SpanContext): Promise<boolean>;
+    writeLargeObject(obj: any, blobId: string, context: SpanContext): Promise<void>;
```
sklose, line 68: this API is very vague ... what kind of object is this? how is it encoded? what if I want a different encoding? How do I read it back (read only returns a string)?

packages/azure/src/utils/BlobClient.ts
```diff
@@ -14,10 +14,13 @@ import {
...
+import { BlobService, createBlobService, ServiceResponse, common } from "azure-storage";
 import { Span, SpanContext, Tags, Tracer } from "opentracing";
 import { IBlobStorageConfiguration } from "..";
 
+import path = require("path");
```
sklose, line 21: can you make this proper typescript imports? and avoid any

packages/azure/src/utils/BlobClient.ts
```diff
@@ -208,4 +198,179 @@ export class BlobClient implements IRequireInitialization {
...
+                        resolve(result);
+                    }
+                }
+            );
+        });
+    }
+
+    public writeLargeObject(obj: any, blobId: string, context: SpanContext): Promise<void> {
+        const filepath: string = path.join(this.localStoragePath!, "temp.json");
+        fs.writeFileSync(filepath, JSON.stringify(obj));
```
sklose, line 310: don't use sync APIs ... it will block the entire node process

sklose: probably want to delete the file after uploading in a `finally` block as well?

packages/azure/src/index.ts
chandanRIT: can you share some examples of unification?

packages/azure/src/index.ts
chandanRIT: done

packages/azure/src/index.ts
chandanRIT: done

packages/azure/src/utils/BlobClient.ts
chandanRIT: changed it to import *

packages/azure/src/index.ts
chandanRIT: done

packages/azure/src/index.ts
chandanRIT: I havent added the read back as we didnt need it. I can add a TODO for the largeRead() method for now

packages/azure/src/utils/BlobClient.ts
chandanRIT: done

packages/azure/src/index.ts
chandanRIT: Updated: blobSubpath-> blobId, folderSubPath -> folderId, blobId as is.

packages/azure/src/utils/BlobClient.ts
chandanRIT: refactor old tests to the new format

packages/azure/src/index.ts
```diff
@@ -49,3 +51,26 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    readAsText(context: SpanContext, blobId: string): Promise<string>;
+    writeAsLargeText(text: string, blobId: string, context: SpanContext): Promise<void>;
+    // TODO: Add a method for reading large text
+
+    exists(context: SpanContext, blobId: string): Promise<boolean>;
+    deleteFolderIfExists(folderId: string, context: SpanContext): Promise<boolean>;
+    deleteBlobIfExists(blobId: string, context: SpanContext): Promise<boolean>;
+    listAllBlobs(
+        prefix: string,
+        continuationToken: IBlobClientPaginationToken,
```
sklose, line 68: how does this `IBlobClientPaginationToken` work? I don't see any function returning one. What is the end-user supposed to pass in here?

chandanRIT: good catch, I shouldnt exposing pagination for listAllBlobs(). Will fix it

chandanRIT: fixed

packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -73,7 +83,7 @@ describe("BlobClient", () => {
         const exists = { exists: true };
 
         beforeEach(() => {
-            MockCreateBlobService.mockImplementation(() => {
```
chandanRIT, line 76: @sklose , I am planning to refactor the old tests similar to how I have the new tests. 

i.e a describe block to test each method, and each describe block will test for success and failure cases of the method under test.

sklose: not sure if you have the time for it, but I would prefer if these tests were actual integration tests and not using a mock underneath. my experience with tests against mocks, where the mock replaces 99% of the thing we wanna test, is that they never uncover any regressions and aren't really useful. we recently added the azure storage simulator to our CI so you can actually hit a real blob storage endpoint (cc @plameniv probably has the details for the endpoint and how to connect). if you add "[full ci]" to your commit message then it will run all the integration tests as well (we currently have them for cosmos only).

packages/azure/src/__test__/utils/BlobClient.test.ts
```diff
@@ -114,4 +124,221 @@ describe("BlobClient", () => {
             await expect(blobClient.exists(span.context(), "BlobID")).resolves.toEqual(true);
         });
     });
+
+    describe("Tests writeAsText()", () => {
+        test("writeAsText() performs a successful write", async () => {
```
sklose, line 129: I personally prefer the `it` style where you make each test case read like a sentence ... `it("writes plain text")`, `it("serializes and writes objects")`, `it("writes binary data"`) ... etc. `describe` ideally groups the test cases into logical units ... `describe(`blob write operations`)

packages/azure/src/index.ts
```diff
@@ -49,3 +51,22 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
     configuration = config.parse(CosmosConfiguration, configuration);
     return new CosmosClient(configuration);
 }
+
+export interface IBlobClient extends IRequireInitialization {
+    createContainerIfNotExists(context?: SpanContext): Promise<BlobService.ContainerResult>;
+
+    writeAsText(context: SpanContext, text: Buffer | string, blobId: string): Promise<void>;
```
sklose, line 58: can you make the order for SpanContext consistent here? either always first, or always last.

packages/azure/src/index.ts
```diff
@@ -49,3 +51,22 @@ export function cosmosQueryClient(configuration: ICosmosConfiguration): ICosmosQ
...
+    readAsText(context: SpanContext, blobId: string): Promise<string>;
+    writeAsLargeText(text: string, blobId: string, context: SpanContext): Promise<void>;
```
sklose, line 60: what's the difference between `writeAsLargeText` and `writeAsText`? why do I have to choose as an end-user? Both accept a string to be written

chandanRIT: Oh I see. I can just expose writeAsText() and depending on the size of the string, I can handle things differently.

packages/azure/src/index.ts
chandanRIT: yes

packages/azure/src/__test__/utils/BlobClient.test.ts
chandanRIT: @sklose I am assuming I can go ahead with the refactoring then.

packages/azure/src/__test__/utils/BlobClient.test.ts
chandanRIT: will change it to it for consistency

packages/azure/src/index.ts
chandanRIT: one downside to having just a single method for writing text however is that the clients will always need to provide the localStoragePath parameter for IBlobStorageConfiguration. 

So maybe we should move writeAsLargeText() into its own class LargeBlobClient, then there wont any confusion.

sklose: I would suggest the following, rename the method to just `write(context: SpanContext, blobId: string, value: Buffer | string)` (you also accept buffers, so it isn't just text). And then internally if the localStoragePath is set AND either the buffer or the string is longer than X, then do the optimized upload version (I assume it would also work with binary data and not just text if there is a larger Buffer). Maybe we can also find a better name for `localStoragePath` that describes its purpose better ... maybe `largeContentTempPath` or something like that?

packages/azure/src/__test__/utils/BlobClient.test.ts
chandanRIT: sure, I can check with Plamen to get started on Integration tests.

packages/azure/src/index.ts
chandanRIT: I can check if the blob library also accepts large buffers. 

and if the user forgets to set the storage path, then we could throw an exception

PR #218, opened 2021-03-10
upgrading gcp dependencies in package

brett-kepler-walmart:
As part of the effort to fix ongoing issues with `task-response-uploader-gcp` service will try to see if upgrading the dependencies can help with some of said issues.

### DIFF
--- a/packages/gcp/package.json
+++ b/packages/gcp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-gcp",
-    "version": "1.4.0-beta.1",
+    "version": "1.4.0-beta.2",
     "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


packages/gcp/package.json
```diff
@@ -13,9 +13,9 @@
         "url": "https://github.com/walmartlabs/cookie-cutter/issues"
     },
     "dependencies": {
-        "@google-cloud/bigquery": "4.7.0",
-        "@google-cloud/pubsub": "2.0.0",
-        "@google-cloud/storage": "4.4.0"
+        "@google-cloud/bigquery": "5.5.0",
```
sklose, line 16: can you bump the version number of the package as well? otherwise, it's not getting published out to the npm registry

Issue #211, opened 2021-01-22
Fix timing metrics returned by gRPC package


k-gupta:
The time being written by the gRPC package is in millisecond - https://github.com/walmartlabs/cookie-cutter/blob/master/packages/grpc/src/internal/client.ts#L301

The base unit for time in Prometheus is seconds. The return value needs to be changed to account for this.

PR #220, opened 2021-03-12
Fixed issue with wrong grpcClient timing metric's unit

hienqnguyen:
This PR is attempting to fix issue #211 where the performance timing being passed in is in millisecs whereas the underlying metric collector expects the unit to be seconds.

I'm not sure if the fix is too restrictive in a sense that we might not be able to instrument in millisecs if we wanted to later on. Perhaps, a better fix is in the following `timing` function?

https://github.com/walmartlabs/cookie-cutter/blob/52aaa98ea418c2e8be923f574c5d19ae23fbd563/packages/prometheus/src/index.ts#L142

### DIFF
--- a/packages/grpc/src/internal/client.ts
+++ b/packages/grpc/src/internal/client.ts
@@ -298,7 +298,7 @@ function emitTimerMetric(
     metrics: IMetrics
 ): void {
     const currentPerformanceTime = performance.now();
-    const runTime = Math.round((currentPerformanceTime - startTime)/1000);
+    const runTime = Math.round((currentPerformanceTime - startTime) / 1000);
     metrics.timing(GrpcMetrics.RequestProcessingTime, runTime, {
         path,
         endpoint,
--- a/packages/grpc/src/internal/client.ts
+++ b/packages/grpc/src/internal/client.ts
@@ -298,7 +298,7 @@ function emitTimerMetric(
     metrics: IMetrics
 ): void {
     const currentPerformanceTime = performance.now();
-    const runTime = Math.round((currentPerformanceTime - startTime) / 1000);
+    const runTime = (currentPerformanceTime - startTime) / 1000;
     metrics.timing(GrpcMetrics.RequestProcessingTime, runTime, {
         path,
         endpoint,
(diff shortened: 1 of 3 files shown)


packages/grpc/src/internal/client.ts
```diff
@@ -298,7 +298,7 @@ function emitTimerMetric(
     metrics: IMetrics
 ): void {
     const currentPerformanceTime = performance.now();
-    const runTime = Math.round(currentPerformanceTime - startTime);
+    const runTime = Math.round((currentPerformanceTime - startTime) / 1000);
```
sklose, line 301: the `Math.round` should probably be removed if you change it to seconds. Otherwise, you only get whole seconds without fractions.

sklose: good question ... the CC convention for `timing` has been milliseconds, I am not sure what all we are breaking with this change. AP might be emitting custom `timing` metrics and already using seconds, so we wouldn't wanna divide those by 1,000 (cc @chrnola maybe you can confirm). On the other hand, doing the conversion in the Prometheus module makes more sense to me in the sense that CC operates in millis, and then the Prometheus module converts it to what's native for Prometheus.

Also are there any other places in CC that emit `timing` metrics? Is anybody currently relying on those being in milliseconds on dashboards?

chrnola: @sklose yeah good point, my team has instrumented a lot of code already assuming that the duration value passed to `timing` will be interpreted verbatim (i.e. we explicitly convert to seconds first).

sklose: @hienqnguyen maybe let's have a look if there are any other `timing` metrics in CC. If the one in `grpc` is the only one then we can maybe just change the spec of the `timing` API and say it expects seconds. Don't think we ever explicitly specified that anywhere to begin with.

hienqnguyen: @sklose as far as I can see.  The only other place is also in `grpc` package here https://github.com/walmartlabs/cookie-cutter/blob/52aaa98ea418c2e8be923f574c5d19ae23fbd563/packages/grpc/src/internal/GrpcInputSource.ts#L190

Will update PR to do conversion there too.

sklose (approved):

PR #229, opened 2021-05-13
Log out a detailed error msg while inserting data into big query

snarayanan1095:
While inserting data into big query, it sends a response back if there's an error. This can be extremely helpful while debugging

### DIFF
--- a/packages/gcp/src/BigQueryClient.ts
+++ b/packages/gcp/src/BigQueryClient.ts
@@ -89,7 +89,7 @@ export class BigQueryClient implements IBigQueryClient, IRequireInitialization {
             });
             let detailedErrorMsg = e;
             if (e.errors) {
-                detailedErrorMsg = JSON.stringify(e.errors);
+                detailedErrorMsg = e.errors;
             }
             this.logger.error("An error occurred while inserting data into big query", {
                 errorMsg: detailedErrorMsg,
--- a/packages/gcp/src/BigQueryClient.ts
+++ b/packages/gcp/src/BigQueryClient.ts
@@ -92,7 +92,7 @@ export class BigQueryClient implements IBigQueryClient, IRequireInitialization {
                 detailedErrorMsg = e.errors;
             }
             this.logger.error(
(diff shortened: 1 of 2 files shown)


packages/gcp/src/BigQueryClient.ts
```diff
@@ -13,6 +13,8 @@ import {
     IMetrics,
     IRequireInitialization,
     OpenTracingTagKeys,
+    ILogger,
+    NullLogger
```
plameniv, line 17: Build is complaining about a missing comma here.

snarayanan1095: Added the comma

packages/gcp/src/BigQueryClient.ts
```diff
@@ -83,6 +88,13 @@ export class BigQueryClient implements IBigQueryClient, IRequireInitialization {
                 result: BigQueryMetricResults.Error,
                 error: e instanceof Error ? e.message : "",
             });
+            let detailedErrorMsg = "";
+            if (e.errors) {
+                detailedErrorMsg = JSON.stringify(e.errors);
+            }
```
plameniv, line 94: Should you have an `else` which stringifies `e`? That way you won't get an empty `errorMsg`.

packages/gcp/src/BigQueryClient.ts
```diff
@@ -47,11 +50,13 @@ export class BigQueryClient implements IBigQueryClient, IRequireInitialization {
             },
         }).dataset(this.config.datasetId);
         this.tracer = DefaultComponentContext.tracer;
+        this.logger = new NullLogger();
```
plameniv, line 53: You can replace `new NullLogger();` with `DefaultComponentContext.logger` and remove the `NullLogger` from the import statement.

packages/gcp/src/BigQueryClient.ts
snarayanan1095: Yeah, will instead make detailedErrorMsg = `e`

plameniv (approved):

PR #230, opened 2021-05-13
Add TLS support to Kafka package

plameniv:


packages/kafka/src/config.ts
```diff
@@ -43,6 +44,14 @@ export class KafkaBrokerConfiguration implements IKafkaBrokerConfiguration {
     public get headerNames(): IKafkaHeaderNames {
         return config.noop();
     }
+
+    @config.field(config.converters.none)
+    public set ssl(_: tls.ConnectionOptions) {
+        config.noop();
+    }
+    public get ssl(): tls.ConnectionOptions {
```
sklose, line 52: are we leaking abstraction here by exporting from "tls"? how complex is ConnectionOptions? Can we redefine the interface and do a mapping?

plameniv: Kafkajs uses the standard library's `tls`. From their docs: `The options are passed directly to tls.connect and used to create the TLS Secure Context, all options are accepted.`. There are 27 unique fields in the `ConnnectionOptions` interface, but I don't know if we want to be the ones deciding what should be used or not.

sklose (approved):

PR #233, opened 2021-05-18
Handle url separately from connectionString

plameniv:


packages/azure/src/index.ts
```diff
@@ -29,6 +29,8 @@ export interface IBlobStorageConfiguration {
     readonly storageAccount: string;
     readonly storageAccessKey: string;
     readonly container: string;
+    readonly connectionString?: string;
```
talam, line 32: can you write up an example of a connectionString?

packages/azure/src/utils/QueueClient.ts
```diff
@@ -135,19 +135,25 @@ export class QueueClient implements IRequireInitialization {
             };
         }
 
-        if (config.url) {
+        const sharedKeyCredential = new StorageSharedKeyCredential(
+            config.storageAccount,
+            config.storageAccessKey
+        );
+
+        if (config.connectionString) {
```
talam, line 143: can you write a doc comment somewhere in the config that notes there preference order for how the client is setup?

talam (approved):

sklose (approved):

PR #235, opened 2021-05-21
Updating version of mssql to recent

brett-kepler-walmart:
Updating version of mssql from 4.3.9 to 7.1.0 to try and debug issue that is happening with sql connections

### DIFF
--- a/packages/mssql/src/index.ts
+++ b/packages/mssql/src/index.ts
@@ -22,6 +22,7 @@ export interface IMssqlConfiguration {
     readonly mode?: Mode;
     readonly connectionTimeout?: number;
     readonly requestTimeout?: number;
+    readonly stream?: boolean;
 }
 
 @config.section
@@ -87,6 +88,14 @@ class MssqlConfiguration {
         config.noop();
     }
     public get requestTimeout(): number {
+        return config.noop();
+    }
+
+    @config.field(config.converters.boolean)
+    public set stream(_: boolean) {
+        config.noop();
+    }
+    public get stream(): boolean {
         return config.noop();
     }
 }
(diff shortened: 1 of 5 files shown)


sklose:
can we push a commit here with [full ci] to ensure the integration tests all pass

plameniv:
Brett mentioned that he successfully ran yarn integrate locally. It also looks like there is some connection issue that was the underlying cause of the problem he was seeing. We are looking at changing the timeout to see if that will fix the connection issues, so there may be additional changes on this PR. I will make sure we run [full ci] before the merge.

packages/mssql/src/index.ts
```diff
@@ -71,6 +74,30 @@ class MssqlConfiguration {
...
+    @config.field(config.converters.boolean)
+    public set stream(_: boolean) {
+        config.noop();
+    }
+    public get stream(): boolean {
+        return config.noop();
+    }
 }
 
 export function mssqlSink(configuration: IMssqlConfiguration): IOutputSink<IPublishedMessage> {
```
plameniv, line 103: Do we want to set a default value for stream (or the timeout values), different from the one defined by mssql package? If so, we should add it here

packages/mssql/src/index.ts
```diff
@@ -71,6 +74,30 @@ class MssqlConfiguration {
     public get mode(): Mode {
         return config.noop();
     }
+
+    @config.field(config.converters.number)
```
plameniv, line 78: Can you replace this with `@config.field(config.converters.timespan)`? It converts values into milliseconds, and that's what the unit that mssql accepts?

packages/mssql/src/index.ts
brett-kepler-walmart: I don't think so because these changes will only matters in one place at the moment

plameniv (approved):

PR #236, opened 2021-05-27
Update packages to reduce vulnerabilities

plameniv:


.travis.yml
```diff
@@ -26,8 +26,8 @@ matrix:
           node_js: 10
           stage: smoke
           if: type = cron
-          script: /bin/bash -c 'yarn audit --level critical; [[ $? -ge 16 ]] && exit 1 || exit 0'
-          # script: yarn audit
+          # script: /bin/bash -c 'yarn audit --level critical; [[ $? -ge 16 ]] && exit 1 || exit 0'
```
sklose, line 29: can you remove the commented line? don't think we need that anymore, do we?

sklose (approved):

Issue #237, opened 2021-06-01
Support ability to specify TTLs at item level when storing events to Azure CosmosDB


hienqnguyen:
We store events of different event streams in a single collection and would like to be able to set their TTLs differently.  Container level TTL is not granular enough for us.

Thus, we'd like to be able to specify the TTL at item level when storing an event. Per Azure [documentation](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-time-to-live?tabs=dotnetv2%2Cjavav4#nodejs-set-ttl-item), the item's level TTL will override the one specified at the container level.

plameniv:
Work is complete and feature is deployed

PR #240, opened 2021-06-07
Include `ttl` setting when storing to Azure cosmosDB

hienqnguyen:
This is part 2 of #237.  When a stored message as a `ttl` defined destined to Azure Cosmos DB, include that setting as per their [docs](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-time-to-live?tabs=dotnetv2%2Cjavav4#nodejs-set-ttl-item)


packages/azure/src/utils/index.ts
```diff
@@ -26,6 +26,12 @@ export interface ICosmosMetadata {
     };
 }
 
+export enum CosmosMetadata {
+    // Cosmos item level TTL in seconds.  Note that this is effective only if container level TTL is enabled
+    // See https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-time-to-live?tabs=dotnetv2%2Cjavav4#nodejs-set-ttl-item
+    ttl = "ttl",
```
sklose, line 32: can you make this caps to be consistent with other enums? TTL

packages/azure/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-azure",
-    "version": "1.4.0-beta.3",
+    "version": "1.4.0-beta.4",
```
plameniv, line 3: Another PR was merge that affected the azure package.json, so please rebase against develop and bump the version to `.5`

packages/azure/src/utils/index.ts
```diff
@@ -26,6 +26,12 @@ export interface ICosmosMetadata {
     };
 }
 
+export enum CosmosMetadata {
```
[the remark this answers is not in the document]
hienqnguyen: Yah, good one.  I didn't think about that.  Done.

sklose (approved):

PR #253, opened 2021-06-17
Core application test increase timeout

plameniv:


sklose (approved):

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-core",
-    "version": "1.3.0",
+    "version": "1.3.1",
```
sklose, line 3: let's not bump the version for this, it's just a test change ... not something an enduser would see as a change

PR #260, opened 2021-06-21
Fix gRPC response stream

prachi30:


sklose (changes requested):
the change is equivalent to the original code, this should make no difference

plameniv:
Superseded by #263

PR #263, opened 2021-06-21
Add await before call to pipe.send

plameniv:


packages/grpc/src/internal/GrpcResponseStream.ts
```diff
@@ -30,7 +30,7 @@ export class GrpcResponseStream<T = any> implements IResponseStream<T> {
         }
     }
 
-    public send(response: T): Promise<void> {
-        return this.pipe.send(response);
+    public async send(response: T): Promise<void> {
```
sklose, line 33: these two implementations should be identical 🤔  returning the Promise directly vs. awaiting it and creating yet another Promise that is returned.

PR #275, opened 2021-07-19
Add schema and bulk insert support for Table mode

plameniv:


packages/mssql/src/__test__/mssql.integration.test.ts
```diff
@@ -116,13 +127,39 @@ function testApp(messages: IMessage[]): CancelablePromise<void> {
             mssqlSink({
                 ...config,
                 encrypt: true,
-                mode: Mode.StoredProcedure,
+                mode,
+                schema,
```
sklose, line 131: I would maybe call this parameter `defaultSchema` and in the event handler allow for this

```
    public onPartialObject(msg: SimpleObject, ctx: IDispatchContext): void {
        ctx.logger.info(JSON.stringify(msg));
        ctx.publish(PartialObject, msg, { [MsSqlMetadata.Schema]: "my-schema" });
    }
```

It's more flexible ... allows you to mix diff schemas in one app.

plameniv: Added support for `MsSqlMetadata.Schema`

packages/mssql/src/MssqlSink.ts
```diff
@@ -276,10 +373,14 @@ export class MssqlSink
     }
 
     private async getSprocDetailsFromSql(
+        schema: string,
         sprocName: string,
         request: sql.Request
     ): Promise<ISprocDetails> {
-        const rawSprocDetails = await request.query(this.getSqlQuery(sprocName));
+        const rawSprocDetails = await request.query(this.getSprocSqlQuery(schema, sprocName));
+        if (rawSprocDetails.rowsAffected[0] === 0) {
```
mag415, line 381: Do you need to check if len(rowsAffected) > 1 ?

Same comment below

mag415 (approved):

PR #285, opened 2021-08-14
Expose KafkaJS Client Configurations

hienqnguyen:
Expose `connectionTimeout` and `requestTimeout` configurations to allow consumers to adjust these settings as needed


mag415 (approved):

packages/kafka/src/index.ts
```diff
@@ -91,6 +91,20 @@ export interface IKafkaSubscriptionConfiguration {
     readonly sessionTimeout?: number;
 }
 
+export interface IKafkaClientConfiguration {
```
plameniv, line 94: You need to create a class implementing this interface inside `config.ts` similarly to how `KafkaBrokerConfiguration` implements `IKafkaBrokerConfiguration`.

Then you need to make `KafkaSubscriptionConfiguration` also extends the new class.

packages/kafka/src/KafkaConsumer.ts
```diff
@@ -137,6 +138,8 @@ export class KafkaConsumer implements IRequireInitialization, IDisposable {
             clientId: generateClientId(),
             brokers: Array.isArray(broker) ? broker : [broker],
             ssl,
+            connectionTimeout: this.config.connectionTimeout || 1000, // KafkaJs Default
```
plameniv, line 141: Remove default values from here

packages/kafka/src/index.ts
```diff
@@ -162,7 +176,7 @@ export interface IKafkaMessagePreprocessor {
 }
 
 export function kafkaSource(
-    configuration: IKafkaBrokerConfiguration & IKafkaSubscriptionConfiguration
+    configuration: IKafkaBrokerConfiguration & IKafkaSubscriptionConfiguration & IKafkaClientConfiguration
 ): IInputSource {
     configuration = config.parse(KafkaSubscriptionConfiguration, configuration, {
         consumeTimeout: 50,
```
plameniv, line 182: Add default values here

packages/kafka/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@walmartlabs/cookie-cutter-kafka",
-    "version": "1.4.0-beta.4",
+    "version": "1.4.0-beta.5",
```
plameniv, line 3: Rebase against current develop as it has been updated

plameniv (changes requested):

packages/kafka/package.json
hienqnguyen: @plameniv  rebased but didn't see any changes upstream from walmartlabs:develop branch.

packages/kafka/src/index.ts
hienqnguyen: makes sense.  Updated.

packages/kafka/src/index.ts
```diff
@@ -172,6 +186,8 @@ export function kafkaSource(
         preprocessor: {
             process: (msg) => msg,
         },
+        connectionTimeout: 1000,
+        requestTimeout: 3000,
```
plameniv, line 190: This was `30000` in a previous commit. Which value is the correct one?

PR #293, opened 2021-11-16
GRPC Health check

hienqnguyen:
Just a proof of concept:
* Add 2 grpc endpoints for health checking as per https://github.com/grpc/grpc/blob/master/doc/health-checking.md


packages/grpc/package.json
```diff
@@ -15,7 +15,8 @@
     "dependencies": {
         "change-case": "4.1.1",
         "grpc": "1.24.11",
-        "opentracing": "0.14.4"
+        "opentracing": "0.14.4",
+        "grpc-ts-health-check": "2.0.6"
```
hienqnguyen, line 19: This should probably be a peer dep and let the user include this when it's enabled?

packages/grpc/src/index.ts
```diff
@@ -56,6 +57,7 @@ export interface IGrpcServerConfiguration {
     readonly host?: string;
     readonly definitions: IGrpcServiceDefinition[];
     readonly skipNoStreamingValidation?: boolean;
+    readonly healthCheck?: { [key: string]: HealthCheckResponse.ServingStatus };
```
hienqnguyen, line 60: This is a hack for me to toy with it.  This should probably be exposed via dispatch context, i.e, `ctx.probe`?

plameniv:
No longer relevant

PR #304, opened 2021-12-15
Gcp pubsub

manasatully:
This PR is a WIP.

I have tried my best to add all features by referring to the cookie-cutter doc and reading through other services that have been implemented. If there seems to be something missing or incorrectly handled please provide your feedback :)


packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+                    {
+                        "attibutes": message.attributes,
+                        "eventTime": message.publishTime,
+                    },
+                    data,
+                    span.context(),
+                );
+
+                msg.once("released", async (msg, __, err): Promise<void> => {
+                    if(err) {
```
butorine, line 71: ack the message here

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+*/
+export class PubSubPullSource implements IInputSource {
+    private subscriber: Subscription;
+    private done: boolean = false;
+    private tracer: Tracer;
+    private logger: ILogger;
+    // array to store a batch of messages - # of msgs in array is defined by maxMsgBatchSize
+    private messages: any[] =[];
+
+    constructor(private readonly config: IGcpAuthConfiguration, private readonly subscriptionName: string, private readonly maxMsgBatchSize: number = 10) {
```
butorine, line 17: Move to something like `PubSubSourceConfiguration`

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+
+                msg.once("released", async (msg, __, err): Promise<void> => {
+                    if(err) {
+                        this.logger.error("Unable to release message", err);
+                    }
+                    span.finish();
+                });
+
+                yield msg;
+                await sleep(100);
```
butorine, line 78: Why is this necessary?

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        /*
+        * for the subscriber there is an 'error' event handler that needs to be added. this will log the errors
+        * and then gracefully remove the event 'message' listner and stop the cookie cutter app
+        */
+
+        let message: any;
+        while(!this.done) {
+            while(this.messages.length !== 0) {
+                message = this.messages.pop();
```
butorine, line 53: This will take the last message in the array, so you'd be processing them out of order

butorine: use `.shift()`

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+
+                const span = this.tracer.startSpan("Consuming messages from Google PubSub");
+                const data: IMessage = {
```
butorine, line 56: Message can be proto encoded, or a json message.  You need to detect that and decode appropriately.  See example from azure queue:
https://github.com/walmartlabs/cookie-cutter/blob/4c67bca40b0d7d238774a1c519ba5ac35187b13a/packages/azure/src/streaming/internal/QueueInputSource.ts#L100

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+        this.subscriber.on('message', (message) => {
+            this.messages.push(message);
+        });
+
+        /*
+        * for the subscriber there is an 'error' event handler that needs to be added. this will log the errors
+        * and then gracefully remove the event 'message' listner and stop the cookie cutter app
+        */
+
+        let message: any;
```
k-gupta, line 50: Delete you are not accessing it anywhere else.

packages/gcp/src/PubSubSource.ts
k-gupta: Change this to const

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
+import { PubSub, Subscription } from "@google-cloud/pubsub";
```
k-gupta, line 1: Are you running a linter on this? I think this project has prettier set so you should run that

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,88 @@
...
+import { IComponentContext, IInputSource, IInputSourceContext, ILogger, IMessage, MessageRef, sleep } from "@walmartlabs/cookie-cutter-core"
```
k-gupta, line 2: Tests are missing can you add them?

packages/gcp/src/PubSubSource.ts
manasatully: removing this

packages/gcp/src/PubSubSource.ts
manasatully: I agree that making message declared as a 'const' is good as the value it holds is not being modified in any way. The reason I chose to have the variable declared using 'let' was to have the variable declaration be a one time step instead of it being created and destroyed multiple times inside the while loop.

packages/gcp/src/PubSubSource.ts
manasatully: I am only focused right now on getting the functionality and test cases setup. Before the final PR will pass it through a linter to meet the prettier guidelines

packages/gcp/src/PubSubSource.ts
manasatully: what do you mean by delete here?

packages/gcp/src/PubSubSource.ts
manasatully: just ignore my previous comment

packages/gcp/src/PubSubSource.ts
manasatully: I am assuming the delete here is to remove the eventlistener...

packages/gcp/src/PubSubSource.ts
k-gupta: One thing you can do is setup your IDE to do it on save. Upto you but makes life a bit easier. 
https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

manasatully: done!

packages/gcp/src/PubSubSource.ts
```diff
@@ -1,30 +1,53 @@
...
-    constructor(private readonly config: IGcpAuthConfiguration, private readonly subscriptionName: string, private readonly maxMsgBatchSize: number = 10) {
-        /*
-        * need to include - error handling if subscription name is incorrect
-        * The subscriber is only allowed to process msgBatchSize messages in a single batch
-        * The service account is authenticated by feeding in the credentials or by putting the json cred file in the environment set to GOOGLE_APPLICATION_CREDENTIALS
-        */
+    private messages: any[] = [];
+
+    constructor(private readonly config: IGcpAuthConfiguration & IPubSubSubscriberConfiguration) {
+        this.config.maxMsgBatchSize = this.config.maxMsgBatchSize || this.MAX_MSG_BATCH_SIZE;
```
butorine, line 40: Don't modify the config that is passed

packages/gcp/src/PubSubSource.ts
```diff
@@ -1,30 +1,53 @@
...
             projectId: this.config.projectId,
-        }).subscription(this.subscriptionName, {
+            credentials: {
+                client_email: this.config.clientEmail,
+                private_key: this.config.privateKey,
+            },
+        }).subscription(this.config.subscriptionName, {
             flowControl: {
-                maxMessages: this.maxMsgBatchSize,
+                maxMessages: this.config.maxMsgBatchSize,
```
butorine, line 50: Instead of modifying the config in constructor, just do the same thing here:
```maxMessages: this.config.maxMsgBatchSize || this.MAX_MSG_BATCH_SIZE```

File: packages/gcp/src/PubSubSource.ts [hunk 11]
```diff
@@ -34,55 +57,92 @@ export class PubSubPullSource implements IInputSource {
...
-        /*
-        * reads max of maxMsgBatchSize from the subsc
-        */
-        this.subscriber.on('message', (message) => {
-            this.messages.push(message);
+    public async *start(): AsyncIterableIterator<MessageRef> {
+        this.subscriber.on("message", (message) => this.messages.push(message));
+        // Error codes - https://cloud.google.com/pubsub/docs/reference/error-codes
+        this.subscriber.on("error", async (error) => {
+            this.logger.error("Subscriber ran into a error", error);
```
butorine, line 64: This won't propagate the error up, right?  I think you should re-throw it to make sure it's handled up the chain by the the application.

packages/gcp/src/PubSubSource.ts
```diff
@@ -34,55 +57,92 @@ export class PubSubPullSource implements IInputSource {
...
-                );
+                const event_type = attributes[EventSourcedMetadata.EventType];
 
-                msg.once("released", async (msg, __, err): Promise<void> => {
-                    if(err) {
-                        this.logger.error("Unable to release message", err);
-                    }
-                    span.finish();
+                let protoOrJsonPayload = data;
+                if(!isEmbeddable(this.config.encoder) && data.type && data.type == "Buffer" && isArray(data.dat)) {
```
butorine, line 80: Is ```data.dat``` a typo?

packages/gcp/src/PubSubSource.ts
```diff
@@ -34,55 +57,92 @@ export class PubSubPullSource implements IInputSource {
...
+
+                        } finally {
+                            span.finish();
+                        }
+                    }
+                );
+
+                yield msgRef;
             }
             await sleep(100);
```
butorine, line 126: I don't think you need this sleep

packages/gcp/src/index.ts
```diff
@@ -50,6 +50,12 @@ export interface IPubSubPublisherConfiguration {
     readonly maxPayloadSize?: number;
 }
 
+export interface IPubSubSubscriberConfiguration {
+    readonly encoder: IMessageEncoder;
+    readonly subscriptionName: string;
+    maxMsgBatchSize?: number;
```
butorine, line 56: This should also be readonly

packages/gcp/src/PubSubSource.ts
```diff
@@ -59,25 +62,30 @@ export class PubSubPullSource implements IInputSource {
...
             while (this.messages.length !== 0) {
                 const message: any = this.messages.shift();
 
-                const { attributes, data } = message as {
-                    attributes: any;
-                    data: IBufferToJSON | any;
-                };
+                const { attributes, data } = this.config.preprocessor
+                    ? this.config.preprocessor.process(message)
+                    : (message as {
```
butorine, line 75: You can change this to
```
: message as IPubSubMessage;
```

packages/gcp/src/model.ts
```diff
@@ -1,17 +1,17 @@
 export interface IRawPubSubMessage {
```
butorine, line 1: Why is this needed?

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+    ILogger,
+    IMetadata,
+    IRequireInitialization,
+    isEmbeddable,
+    MessageRef,
+    sleep,
+} from "@walmartlabs/cookie-cutter-core";
+import { IGcpAuthConfiguration, IPubSubSubscriberConfiguration } from ".";
+import { FORMAT_HTTP_HEADERS, Tracer, Tags } from "opentracing";
+import { AttributeNames } from "./PubSubSink";
```
plameniv, line 16: Move `AttributeNames` in a `model.ts` file. Both `PubSubSink` and `PubSubSource` should import this const from the file.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+
+interface IBufferToJSON {
```
plameniv, line 18: Move this interface to the `index.ts` file and export it there.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+    type: string;
+    data: any[];
+}
+
+export enum GooglePubSubTracingTagKeys {
```
plameniv, line 23: I don't think this needs to be exported. It is only used in this file.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+    SubscriptionName = "google.pubsub.subscription_name",
+}
+
+export interface IPubSubMessage {
```
plameniv, line 27: Move this interface to the `index.ts` file and export it there.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+
+                yield msgRef;
+            }
+            await sleep(100);
+        }
+    }
+
+    public async stop(): Promise<void> {
+        this.done = true;
+        await this.dispose();
```
plameniv, line 139: Do not dispose here. The application calls the dispose at a later step.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public async dispose(): Promise<void> {
+        if (this.subscriber) {
+            this.subscriber.removeAllListeners();
+            await this.subscriber.close();
+        }
+    }
+
+    private decode(payload: any, event_type: string) {
```
plameniv, line 149: Add the return type, even if it's `any`

packages/gcp/src/config.ts
```diff
@@ -153,3 +155,39 @@ export class PubSubPublisherConfiguration extends GcpAuthConfiguration
...
+    implements IPubSubSubscriberConfiguration {
+    @config.field(config.converters.none)
+    public set encoder(_: IMessageEncoder) {
+        config.noop();
+    }
+    public get encoder(): IMessageEncoder {
+        return config.noop();
+    }
+
+    @config.field(config.converters.none)
```
plameniv, line 170: Use `config.converters.string`

packages/gcp/src/config.ts
```diff
@@ -153,3 +155,39 @@ export class PubSubPublisherConfiguration extends GcpAuthConfiguration
...
+    public set subscriptionName(_: string) {
+        config.noop();
+    }
+    public get subscriptionName() {
```
plameniv, line 174: Add return type for all `get` functions.

packages/gcp/src/index.ts
```diff
@@ -103,3 +117,18 @@ export function pubSubSink(
     });
     return new PubSubSink(configuration);
 }
+
+export interface IPubSubMessagePreprocessor {
```
plameniv, line 121: Move this interface up with the other interfaces. And move `const MAX_MSG_BATCH_SIZE_SUBSCRIBER` up before the interfaces.

packages/gcp/src/index.ts
```diff
@@ -103,3 +117,18 @@ export function pubSubSink(
...
+    process(payload: string): IPubSubMessage;
```
plameniv, line 122: `IPubSubMessagePreprocessor`'s nested interfaces should be exported so they are available to users of this package.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+        this.logger = context.logger;
+    }
+
+    public async *start(): AsyncIterableIterator<MessageRef> {
+        this.subscriber.on("message", (message) => this.messages.push(message));
+        this.subscriber.on("error", (error) => {
+            throw error;
+        });
+
+        while (!this.done) {
```
plameniv, line 68: The current loop in a loop may get in a situation where stop has been called, but you keep having messages in the array. The source will process an unknown number of messages before exiting the inner loop and doing the outer loop check even though you want to stop processing immediately.

Instead do this:
```
for await (const item of this.messages) {
    if (this.done) {
        break;
    }
```
[Example](https://github.com/walmartlabs/cookie-cutter/blob/develop/packages/core/src/defaults/StaticInputSource.ts#L26)

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+                    protoOrJsonPayload = data.data;
+                }
+
+                const msg = this.decode(protoOrJsonPayload, event_type);
+
+                const spanContext = this.tracer.extract(FORMAT_HTTP_HEADERS, attributes);
+                const span = this.tracer.startSpan("Processing Google PubSub Message", {
+                    childOf: spanContext,
+                });
+                span.setTag(Tags.SPAN_KIND, Tags.SPAN_KIND_MESSAGING_CONSUMER);
```
plameniv, line 94: For consistency, can you create a `spanLogAndSetTags` function similar to the one in `PubSubSink`.

You may also need to add `span.log({ topic });` and any other relevant information.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+                msgRef.once(
+                    "released",
+                    async (_msg: MessageRef, _value?: any, error?: Error): Promise<void> => {
+                        try {
+                            if (error) {
+                                this.logger.error(`Unable to release message | ${error}`);
+                            } else {
+                                message.ack();
+                                this.logger.debug(`Message processed : ${message.id}`);
+                                failSpan(span, error);
```
plameniv, line 123: This `failSpan` should be in the `if` part.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+                const msgRef = new MessageRef(metadata, msg, span.context());
+
+                msgRef.once(
+                    "released",
+                    async (_msg: MessageRef, _value?: any, error?: Error): Promise<void> => {
+                        try {
+                            if (error) {
+                                this.logger.error(`Unable to release message | ${error}`);
+                            } else {
+                                message.ack();
```
plameniv, line 121: Not sure how this should be handled, but our handling of message ack usually involves a `catch` statement that throws the caught error. [Amqp Example](https://github.com/walmartlabs/cookie-cutter/blob/develop/packages/amqp/src/AmqpSource.ts#L105), [Redis Example](https://github.com/walmartlabs/cookie-cutter/blob/develop/packages/redis/src/RedisStreamSource.ts#L224).

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    ParallelismMode,
+    sleep,
+} from "@walmartlabs/cookie-cutter-core";
+import {
+    IGcpAuthConfiguration,
+    IPubSubMessagePreprocessor,
+    IPubSubSubscriberConfiguration,
+    MAX_MSG_BATCH_SIZE_SUBSCRIBER,
+    pubSubSource,
+} from "../index";
```
plameniv, line 19: Can replace with `".."`

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+            parallelism: {
+                mode: ParallelismMode.Serial,
+            },
+        });
+}
+
+describe("Testing of pubsub subscriber service WITH DEFAULT batch size", () => {
+    let testConfig: IGcpAuthConfiguration & IPubSubSubscriberConfiguration;
+    beforeEach(() => {
+        testConfig = {
```
plameniv, line 76: `testConfig` never changes between tests, so you can move it out of the `beforeEach` and make it a `const`

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+            if (event === "error") {
+                callback(new Error("This is a test error"));
+            }
+        };
+        const source = pubSubSource({ ...testConfig });
+        const testApp = createTestApp(source);
+        await expect(testApp).rejects.toThrow();
+    });
+
+    it("verifying 'message' event listner gets messages and sends downstream", async () => {
```
plameniv, line 97: `listner` -> `listener` in 3 locations.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,158 @@
...
+                                this.logger.debug(`Message processed : ${message.id}`);
```
plameniv, line 122: Would it be useful to log any other information from the message?

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+            projectId: "test_project_id",
+            clientEmail: "test@testserver.com",
+            privateKey: "test_private_key",
+            encoder: new JsonMessageEncoder(),
+            subscriptionName: "test_subscription_name",
+        };
+        capturedOutput = [];
+    });
+
+    it("Verifying 'error' event listener", async () => {
```
plameniv, line 86: The `it("...")` should read like a sentence that starts with `it`: `it("verifies 'error' event listener",...`. Fix all the it string to follow that format.

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+            clientEmail: "test@testserver.com",
+            privateKey: "test_private_key",
+            encoder: new JsonMessageEncoder(),
+            subscriptionName: "test_subscription_name",
+            maxMsgBatchSize: 10,
+        };
+        capturedOutput = [];
+    });
+
+    it("Verifying 'error' event listener", async () => {
```
plameniv, line 143: Do you really need this test in every `describe`?

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+            await sleep(50);
+        }
+
+        testApp.cancel();
+        await testApp;
+        expect(capturedOutput.length).toBe(testConfig.maxMsgBatchSize);
+    });
+});
+
+describe("Testing with message preprocessor mentioned", () => {
```
plameniv, line 186: remove the word `mentioned`

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    let testConfig: IGcpAuthConfiguration & IPubSubSubscriberConfiguration;
+    function pubSubMessagePreprocessor(): IPubSubMessagePreprocessor {
+        return {
+            process(payload: string): IPubSubMessage {
```
plameniv, line 190: Looks like this would take the entire `message` and place it in the data field of the returned message. Can you create a better example and then check one of the output messages for correctness?

package.json
```diff
@@ -42,3 +42,4 @@
     "bump-version": "node bump-version.js"
   }
 }
+
```
plameniv, line 45: Remove line

plameniv:
In `package.json` for the package, increment the patch version.

packages/gcp/src/PubSubSource.ts
manasatully: The ack function does not throw any errors from what I gather from the docs and other sources. The best way to handle this is to have a listener that subscribes to 'error' events and handle them accordingly. I have a handler to do just that. Instead of handling the error in the callback function I'm throwing it so CC will handle it. Maybe this [https://github.com/googleapis/nodejs-pubsub/issues/575](url) will give you more info.

packages/gcp/src/index.ts
manasatully: The IPubSubMessage interface has been declared and exported [https://github.com/manasatully/cookie-cutter/blob/c98ff95268b16a29c5a170947f46e73b34541fd9/packages/gcp/src/PubSubSource.ts#L27](url)

packages/gcp/src/model.ts
```diff
@@ -0,0 +1,5 @@
+export const AttributeNames = {
```
plameniv, line 1: Add the `Copyright (c) Walmart Inc.`  text that you can see in other files at the top

packages/gcp/src/index.ts
```diff
@@ -27,7 +27,9 @@ import {
 import { GcsClient } from "./GcsClient";
 import { GcsSink } from "./GcsSink";
 import { PubSubSink } from "./PubSubSink";
-import { PubSubSource, IPubSubMessage } from "./PubSubSource";
+import { IBufferToJSON, PubSubSource } from "./PubSubSource";
```
plameniv, line 30: Move the `IBufferToJSON` here, too.

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -16,9 +16,9 @@ import {
     IPubSubSubscriberConfiguration,
     MAX_MSG_BATCH_SIZE_SUBSCRIBER,
     pubSubSource,
-} from "../index";
-import { AttributeNames } from "../PubSubSink";
-import { IPubSubMessage } from "../PubSubSource";
+} from "..";
+import { AttributeNames } from "../model";
+import { IPubSubMessage } from "../index";
```
plameniv, line 21: Move this to the ".." import group above.

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -94,7 +94,7 @@ describe("Testing of pubsub subscriber service WITH DEFAULT batch size", () => {
         await expect(testApp).rejects.toThrow();
     });
 
-    it("verifying 'message' event listner gets messages and sends downstream", async () => {
+    it("verifying 'message' event listener", async () => {
```
plameniv, line 97: `it("verifying` -> `it("verifies`

You can also keep the longer more detailed string. Applies to all the `it(` lines.

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,162 @@
...
+            return this.config.encoder.decode(
+                this.config.encoder.fromJsonEmbedding(payload),
+                event_type
+            );
+        }
+        return this.config.encoder.decode(payload, event_type);
+    }
+
+    private spanLogAndSetTags(span: Span): void {
+        span.setTag(Tags.SPAN_KIND, Tags.SPAN_KIND_MESSAGING_CONSUMER);
```
plameniv, line 153: Add `span.log({ this.config.subscriptionName });`

packages/gcp/src/PubSubSource.ts
```diff
@@ -130,13 +124,12 @@ export class PubSubSource implements IInputSource, IRequireInitialization {
 
                 yield msgRef;
             }
-            await sleep(100);
+            await sleep(50);
```
plameniv, line 127: Is this needed?

packages/gcp/src/__test__/PubSubSource.test.ts
```diff
@@ -0,0 +1,244 @@
...
+import {
+    Application,
+    CancelablePromise,
+    CapturingOutputSink,
+    ConsoleLogger,
+    ErrorHandlingMode,
+    IDispatchContext,
+    IInputSource,
+    JsonMessageEncoder,
+    // ParallelismMode,
```
plameniv, line 17: Remove comment

packages/gcp/src/PubSubSource.ts
```diff
@@ -0,0 +1,177 @@
...
+ * implements pull delivery with limits on max number of unacknowledged messages
+ * that a subscriber can have in process while reading from a topic
+ */
+export class PubSubSource implements IInputSource, IRequireInitialization {
+    private subscriber: Subscription;
+    private done: boolean = false;
+    private tracer: Tracer;
+    private logger: ILogger;
+    private metrics: IMetrics;
+    private queue: BoundedPriorityQueue<any>;
```
plameniv, line 43: Switch the type of the Queue MessageRef instead of any

plameniv (approved):

Issue #324, opened 2022-02-18
After Upgrading azure package to 1.4.0, QueueInputSource failed to read messages written by older (<1.4.0) CC QueueOutputSink with default preprocessor


YashGajera:
It appears that the latest 1.4.0 azure package is using [@azure/storage-queue](https://www.npmjs.com/package/@azure/storage-queue) lib instead of [azure-storage](https://www.npmjs.com/package/azure-storage) lib used by the older CC version. Apparently, there is an inconsistency between how messages are encoded (by default) and how they are decoded between [@azure/storage-queue](https://www.npmjs.com/package/@azure/storage-queue) and [azure-storage](https://www.npmjs.com/package/azure-storage). Newer library failed to decode message encoded by default TextXmlQueueMessageEncoder with older library.

Error: 
`SyntaxError: Unexpected token & in JSON at position 1
    at JSON.parse (<anonymous>)
    at EnvelopeQueueMessagePreprocessor.process (/Users/yug0003/Documents/store-initialization-checkpoints/services/store-initialization-checkpoints/node_modules/@walmartlabs/cookie-cutter-azure/dist/utils/QueueClient.js:43:21)
    at /Users/yug0003/Documents/store-initialization-checkpoints/services/store-initialization-checkpoints/node_modules/@walmartlabs/cookie-cutter-azure/dist/utils/QueueClient.js:221:56
    at Array.reduce (<anonymous>)
    at QueueClient.<anonymous> (/Users/yug0003/Documents/store-initialization-checkpoints/services/store-initialization-checkpoints/node_modules/@walmartlabs/cookie-cutter-azure/dist/utils/QueueClient.js:219:57)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/yug0003/Documents/store-initialization-checkpoints/services/store-initialization-checkpoints/node_modules/@walmartlabs/cookie-cutter-azure/dist/utils/QueueClient.js:11:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)`

Encoded message written with older CC azure queue sink:
[...]

PR #330, opened 2022-04-05
Bug/fix  - issue 324 

YashGajera:
Issue link: https://github.com/walmartlabs/cookie-cutter/issues/324
Added JSON text xml characters decoding logic

### DIFF
--- a/packages/azure/src/__test__/utils/QueueClient.test.ts
+++ b/packages/azure/src/__test__/utils/QueueClient.test.ts
@@ -91,6 +91,16 @@ describe("QueueClient", () => {
                     dequeueCount: 1,
                     messageText: '{ "headers": {"event_type": "event"}, "payload":"data"}',
                 },
+                {
+                    messageId: 124,
+                    insertedOn: new Date(),
+                    expiresOn: new Date(),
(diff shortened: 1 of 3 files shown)


packages/azure/src/__test__/utils/QueueClient.test.ts
```diff
@@ -211,8 +221,9 @@ describe("QueueClient", () => {
         const client = new QueueClient(configuration);
         it("should read messages with defaults", async () => {
             const messages = await client.read(span.context());
-            expect(messages).toHaveLength(1);
+            expect(messages).toHaveLength(2);
             expect(messages[0].headers[QueueMetadata.MessageId]).toBe(123);
+            expect(messages[1].headers[QueueMetadata.MessageId]).toBe(124);
```
plameniv, line 226: Can you add and expect for each message that checks if the decoded value is `{ "headers": {"event_type": "event"}, "payload":"data"}` for both?

YashGajera: Added test to match payload as well.

plameniv (approved):

PR #338, opened 2022-04-27
Add Node 16 Build Step

plameniv:


cross311 (approved):
looks like we are testing everything on 12 and usign 16 which makes sense

plameniv:
Switching the minimum Node version from `10` to `12`.

The Smoke step as well as all Integration steps go from Node `10` to `16`.

Will keep a `Ubuntu Node 12` build step for `Full CI` builds for now, so we still test against `12` while people transition to `16`. 
This step will be changed in the near future to instead look forward and build Node `18`.

PR #342, opened 2022-05-16
Reduce cardinality GRPC peer remove from metrics

cross311:
Peer is creating a bunch of cardinality aspects in metrics.  More of a logging thing rather then a metrics.

### DIFF
--- a/packages/grpc/src/internal/GrpcInputSource.ts
+++ b/packages/grpc/src/internal/GrpcInputSource.ts
@@ -137,7 +137,6 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
 
                     this.metrics.increment(GrpcMetrics.RequestReceived, {
                         path: method.path,
-                        peer: call.getPeer(),
                     });
                     const metadata = call.metadata.getMap();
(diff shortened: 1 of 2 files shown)


plameniv:
Bump the patch version in package.json

plameniv:
Should remove the `peer: call.getPeer()` from the 2 `metrics.increment` calls as well?

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -137,7 +137,6 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
 
                     this.metrics.increment(GrpcMetrics.RequestReceived, {
                         path: method.path,
-                        peer: call.getPeer(),
```
michaelliao5, line 140: what does getPeer do?

michaelliao5 (approved):

packages/grpc/src/internal/GrpcInputSource.ts
plameniv: `Get the endpoint this call/stream is connected to. Returns the URI of the endpoint`

PR #344, opened 2022-06-24
Added ability to pull custom message headers into message metadata for kafka

butorine:
Sometimes messages that were produced by another system have additional information as part of the message headers.  There is a need to propagate that information as part of message metadata.  This PR allows users to specify additional header names that could be added to message metadata.

### DIFF
--- a/packages/kafka/src/__test__/KafkaSource.test.ts
+++ b/packages/kafka/src/__test__/KafkaSource.test.ts
@@ -50,7 +50,11 @@ describe("KafkaSource", () => {
                 offset,
                 partition,
                 key: Buffer.from("key"),
-                headers: { "X-Message-Type": "application/json" },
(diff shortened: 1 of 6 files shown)


packages/kafka/src/__test__/KafkaSource.test.ts
```diff
@@ -70,6 +78,10 @@ describe("KafkaSource", () => {
                 encoder,
                 eos: true,
                 headerNames: DefaultKafkaHeaderNames,
+                additionalHeaderNames: {
+                    header1: additionalHeaderName1,
```
plameniv, line 82: Can you add a few examples of with headers that stay the same (key and value are the same) and some that change?

For example: `internalHeader: “externalHeader”`

plameniv (approved):

PR #346, opened 2022-06-28
added mqtt module

manasatully:


.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
plameniv, line 1: Delete this file and the .yarn directory.

docs/docs/Module_Mqtt.md
```diff
@@ -0,0 +1,116 @@
...
+
+| Name | Description |
+| ---- | ----------- |
+| hostName | MQTT broker's host name |
+| hostPort | MQTT broker's host port where the service is listening on |
+| encoder | defines how the raw data received from MQTT broker should be converted into message objects |
+| topic | name of the subscribed topic where the message was received from |
+| _prepreprocessor_ | optional preprocessing function that allows for manipulation of the incoming message before decoding the message's payload. This can for instance be used for messages that are enveloped in some way |
+| _queueSize_ | the max number of messages that are being processed at a time, default value is 10 |
+| _qos_ | quality of service that can `only` take values 0, 1, and 2. Default value is 0  |
```
plameniv, line 36: What does each number mean?

docs/docs/Module_Mqtt.md
```diff
@@ -0,0 +1,116 @@
...
+
+The available configuration options are
+
+| Name | Description |
+| ---- | ----------- |
+| hostName | MQTT broker's host name |
+| hostPort | MQTT broker's host port where the service is listening on |
+| encoder | defines how the raw data received from MQTT broker should be converted into message objects |
+| topic | name of the subscribed topic where the message was received from |
+| _qos_ | quality of service that can `only` take values 0, 1, and 2. Default value is 0  |
```
plameniv, line 80: What does each number mean?

docs/docs/Module_Mqtt.md
```diff
@@ -0,0 +1,116 @@
...
+    .output()
+        .published(mqttSink({
+            hostName: "test-broker",
+            hostPort: "1234",
+            ecnoder: new JsonMessageEncoder(),
+            topic: "test-topic",
+        }))
+        .done()
+    // ...
+    .run ();
```
plameniv, line 103: remove space between run and ()

packages/mqtt/src/mqttSink.ts
```diff
@@ -0,0 +1,155 @@
+import {
```
plameniv, line 1: Rename file to `MqttSink.ts`

packages/mqtt/src/mqttSource.ts
```diff
@@ -0,0 +1,168 @@
+import {
```
plameniv, line 1: Rename file to `MqttSource.ts`

packages/mqtt/src/mqttSource.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public async initialize(context: IComponentContext): Promise<void> {
+        this.tracer = context.tracer;
+        this.logger = context.logger;
+        this.metrics = context.metrics;
+    }
+
+    public async *start(): AsyncIterableIterator<MessageRef> {
+        this.client.on("connect", (packet: mqtt.IConnackPacket) => {
```
plameniv, line 48: Should the on connect section not be in the `initialize` function?

packages/mqtt/src/mqttSource.ts
```diff
@@ -0,0 +1,168 @@
...
+            this.logger.info("Subscriber made connection to server", {
+                cmd: packet.cmd,
+                returnCode: packet.returnCode,
+                reasonCode: packet.reasonCode,
+            });
+
+            this.client.subscribe(this.config.topic, { qos: this.config.qos });
```
plameniv, line 55: Do we need to unsubscribe in the `dispose` function?

packages/mqtt/src/__test__/MqttSource.test.ts
```diff
@@ -0,0 +1,276 @@
...
+
+describe("Testing mqtt subscriber preprocessor", () => {
+    function mqttMessagePreprocessor(): IMqttPreprocessor {
+        return {
+            process(payload: any): IMqttMessage {
+                const data: any = JSON.parse(Buffer.from(payload).toString());
+                return {
+                    attributes: {
+                        eventType: TestEvent.name,
+                        dt: new Date().getUTCDate(),
```
plameniv, line 186: `getUTCDate` seems to return the day of the month. Is that what you want here?

packages/mqtt/src/mqttSource.ts
```diff
@@ -0,0 +1,168 @@
...
+        });
+
+        this.client.on("message", async (topic: string, payload: Buffer) => {
+            const { attributes, data } = this.config.prepreprocessor
```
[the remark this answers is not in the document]
manasatully: I added the metric for message received. I know that mqtt broker does allow empty/null messages to be sent to the topic not sure of undefined messages though. I don't think that I should handle the case for an empty/null message because what if the user wants an empty or null message to be sent through then I should just pass it forward and not worry about it. That is why there is a message preprocessor. If the user sees that there is a possibility of an empty/null message being received over the topic then it shall be handled by the user with the a message preprocessor. From what I see CC framework expects the message to be of MessageRef format which I am handling its just that the case is an empty payload case.

docs/docs/Module_Mqtt.md
```diff
@@ -47,7 +47,8 @@ The available configuration options are
 
 | Name | Description | Type | Tags |
 | ---- | ----------- | ---- | ---- |
-| cookie_cutter.mqtt_source.msg_received | increases the number of messages received from topic | `increment` | `topic`, `eventType`, `port`, `hostName`, `result`[^1] |
+| cookie_cutter.mqtt_source.msg_received | increases the number of messages received from topic that was processed | `increment` | `topic`, `eventType`, `port`, `hostName`, `result`[^1] |
```
plameniv, line 50: `msg_received` -> `msg_processed`
that was processed -> that were processed

docs/docs/Module_Mqtt.md
```diff
@@ -47,7 +47,8 @@ The available configuration options are
...
+| cookie_cutter.mqtt_source.msg_received | increases the number of messages that is received from the broker | `increment` | `hostName`, `port`, `topic` |
```
plameniv, line 51: that is received -> that were received

packages/mqtt/src/MqttSink.ts
```diff
@@ -0,0 +1,155 @@
...
+    private emitMetrics(topic: string, eventType: string, result: string): void {
+        this.metrics.increment(MqttMetrics.MsgPublished, {
+            topic,
+            eventType,
+            result,
+        });
+    }
+
+    private formattedMessage(message: IPublishedMessage): IPayloadWithAttributes {
+        const timestamp: string = new Date().getUTCDate().toString();
```
plameniv, line 121: `getUTCDate()` returns the day of the month from the date.
You should replace with `Date.now().toString()`

packages/mqtt/src/__test__/MqttSource.test.ts
plameniv: Should replace with `Date.now().toString()`

packages/mqtt/src/MqttSink.ts
```diff
@@ -0,0 +1,155 @@
...
+    implements IOutputSink<IPublishedMessage>, IRequireInitialization, IDisposable {
+    private readonly client: mqtt.MqttClient;
+    private tracer: Tracer;
+    private logger: ILogger;
+    private metrics: IMetrics;
+    private readonly spanOperationName: string = "Publish to MQTT broker";
+    private readonly spanTagComponent: string = "cookie-cutter-mqtt";
+
+    public constructor(private readonly config: IMqttAuthConfig & IMqttPublisherConfiguration) {
+        this.client = mqtt.connect({
```
plameniv, line 38: Looking at this again, I think the `mqtt.connect` should be moved to the initialize. The constructor shouldn't really start the client and given that we have to register listeners, it means the client may be active before we get to the `initialize` call and the listeners that handle connection and possible errors.

You can also add the following in the constructor for all the logger, tracer and metrics: 
`this.logger = DefaultComponentContext.logger;`

packages/mqtt/src/MqttSource.ts
```diff
@@ -0,0 +1,175 @@
...
+    private readonly client: mqtt.Client;
+    private tracer: Tracer;
+    private logger: ILogger;
+    private metrics: IMetrics;
+    private readonly queue: BoundedPriorityQueue<MessageRef>;
+    private readonly spanOperationName: string = "Received MQTT message from broker";
+    private readonly spanTagComponent: string = "cookie-cutter-mqtt";
+
+    public constructor(private readonly config: IMqttAuthConfig & IMqttSubscriberConfiguration) {
+        this.client = mqtt.connect({
```
plameniv, line 33: Same here as well: move to initialize function.

packages/mqtt/src/MqttSink.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    public async sink(output: IterableIterator<IPublishedMessage>): Promise<void> {
+        for (const message of output) {
+            this.client.publish(
+                this.config.topic,
+                Buffer.from(JSON.stringify(message.message)),
+                { qos: this.config.qos },
+                (error: Error) => {
+                    const formattedMsg: IPayloadWithAttributes = this.formattedMessage(message);
```
plameniv, line 69: `formattedMsg` is only used locally to get access to several of its member variables. However you don't need to create this variable to access the underlying data.

Is the formatted Msg supposed to go to MQTT instead of the straight `IMessage` (message.message)?

docs/docs/Module_Mqtt.md
```diff
@@ -0,0 +1,117 @@
...
+
+The available configuration options are
+
+| Name | Description |
+| ---- | ----------- |
+| hostName | MQTT broker's host name |
+| hostPort | MQTT broker's host port where the service is listening on |
+| encoder | defines how the raw data received from MQTT broker should be converted into message objects |
+| topic | name of the subscribed topic where the message was received from |
+| _prepreprocessor_ | optional preprocessing function that allows for manipulation of the incoming message before decoding the message's payload. This can for instance be used for messages that are enveloped in some way |
```
plameniv, line 34: Is this a typo - `prepreprocessor`? It's also in several spots in the code

plameniv (approved):

PR #349, opened 2022-07-12
Fix yarn --json not returning proper json

plameniv:


cross311 (approved):
Was there some change in our yarn version why this is now broke?

plameniv:
The change turned out to be in how `npm show <pkg> --json` outputs json when it encounters an error (in our case a package that's not published). Version `8.11.0` fails to output proper json. I've added a step installing a `6.14.6` which throws the error but still outputs a properly formatted json of the error.

PR #360, opened 2022-08-09
Update from grpc to @grpc/grpc-js

plameniv:


packages/grpc/src/internal/client.ts
```diff
@@ -73,15 +74,32 @@ class ClientBase implements IRequireInitialization, IDisposable {
...
     const serviceDef = createServiceDefinition(config.definition);
+    let client: Client;
     const ClientType = makeGenericClientConstructor(serviceDef, undefined, undefined);
-    const client = new ClientType(
-        config.endpoint,
-        credentials.createInsecure(),
-        createGrpcConfiguration(config)
-    );
+    if (certPath) {
+        const rootCert = readFileSync(certPath);
```
michaelliao5, line 84: will this cert logic flow throw error ?

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -205,24 +216,23 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
...
 
     public async stop(): Promise<void> {
-        await this.queue.close();
-        await this.streamHandler.dispose();
+        this.queue.close();
+    }
 
-        const graceful = new Promise<void>((resolve) => {
-            this.server.tryShutdown(() => {
+    public async dispose(): Promise<void> {
```
michaelliao5, line 222: is dispose function called automatically

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -93,11 +92,22 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
         this.tracer = context.tracer;
         this.metrics = context.metrics;
         await this.streamHandler.initialize(context);
-
-        this.server.bind(
-            `${this.config.host}:${this.config.port}`,
-            ServerCredentials.createInsecure()
-        );
+        await new Promise<void>(async (resolve, reject) => {
+            this.server.bindAsync(
```
michaelliao5, line 96: whats the main reason for changing from sync bind to away async bind

plameniv: New library doesn't implement the bind function, just the bindAsync

packages/grpc/src/internal/GrpcInputSource.ts
plameniv: Dispose is called by the Application on shut down. I had to move things here because they weren't getting called on time via the stop function.

Currently dispose is awaited before the rest of the shut down process completes.

packages/grpc/src/internal/client.ts
plameniv: readFileSync should throw an error if file doesn't exist for example.

michaelliao5 (approved):

PR #369, opened 2022-11-09
Allowing streams for large uploads to google storage via the GCS client

k0p04ku:
**Testing**
image
image

I was able to successfully stream a video into GS.

### DIFF
--- a/packages/gcp/src/__test__/GcsClient.test.ts
+++ b/packages/gcp/src/__test__/GcsClient.test.ts
@@ -35,11 +35,7 @@ describe("GcsClient", () => {
     describe("Proceeds with expected failure", () => {
         const err = "A DEFINED VALUE";
         const content = Buffer.from("CONTENTS TO BE WRITTEN");
-        const stream = new Readable({
-            read(_size) {
-                this.push(err);
-            },
-        });
+        const stream = Readable.from(err);
 
         beforeEach(() => {
             const mockFile = {
--- a/packages/gcp/src/__test__/GcsClient.test.ts
+++ b/packages/gcp/src/__test__/GcsClient.test.ts
@@ -128,7 +128,6 @@ describe("GcsClient", () => {
             await expect(
                 client.putObjectAsStream(span.context(), mockReadStream, "fileName")
             ).resolves.toBe(undefined);
-            await new Promise((resolve) => setTimeout(resolve, 1000));
             expect(mockWriteStream.buffer).toEqual(data);
         });
     });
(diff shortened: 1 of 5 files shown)


cross311 (changes requested):
I think this is a good change, just want to see testing more concrete that it works.

packages/gcp/src/__test__/GcsClient.test.ts
```diff
@@ -86,5 +115,13 @@ describe("GcsClient", () => {
                 undefined
             );
         });
+
+        it("performs a successful write via readable stream", async () => {
+            const client = gcsClient(config);
+            await client.initialize(ctx);
+            await expect(client.putObjectAsStream(span.context(), mockStream, "fileName")).resolves.toBe(
```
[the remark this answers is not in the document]
k0p04ku: I've added an assertion for this now.

k0p04ku:
> I think this is a good change, just want to see testing more concrete that it works.

I've added an assertion regarding what exactly we're reading and writing into GS.

packages/gcp/src/__test__/GcsClient.test.ts
```diff
@@ -34,22 +35,25 @@ describe("GcsClient", () => {
...
+                file: (_) => mockFile,
+            };
+            MockStorage.mockImplementation(() => {
+                return {
+                    bucket: () => mockBucket,
+                };
+            });
         });
 
         it("rejects on error from gcs for put", async () => {
```
prachi30, line 59: nit: can you re word it to be specific to put object?

prachi30:
```suggestion
        it("rejects on error from gcs for putting objects", async () => {
```

k0p04ku: Fixed

prachi30 (approved):

prachi30 (approved):

packages/gcp/src/index.ts
```diff
@@ -78,6 +79,7 @@ export interface IPubSubMessagePreprocessor {
 
 export interface IGcsClient {
     putObject(spanContext: SpanContext, body: Buffer, key: string): Promise<void>;
+    putObjectAsStream(context: SpanContext, stream: Readable, key: string): Promise<void>;
```
cross311, line 82: Mind keeping variable names consistant
```suggestion
    putObjectAsStream(spanContext: SpanContext, stream: Readable, key: string): Promise<void>;
```

cross311 (changes requested):

packages/gcp/src/index.ts
k0p04ku: Fixed.

cross311 (approved):

PR #374, opened 2022-12-28
Add cc health check to prevent stuck apps for any input sources

prachi30:
This health check will enable applications to auto heal/log if input queue of cc is full and no new messages from it are being processed for certain amount of time. It could be used as a guard for apps using kafka, redis, etc. and would be particularly helpful for kafka sources where we have a known issue because of cookie cutter uses sync callback to implement synchronization barrier for post rebalance async calls which sometimes causes consumers to be lost. Kafka JS has an open issue to add support for async callbacks but until that is done, this health check will help. Similarly it would be a good guard to add for apps with redis source as well.

### DIFF
--- a/packages/core/src/__test__/Application.test.ts
+++ b/packages/core/src/__test__/Application.test.ts
@@ -690,7 +690,6 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
                 .published(new CapturingOutputSink(capture))
                 .done()
                 .run(runtime);
-            await Promise.race([sleep(1000), app]);
             app.cancel();
             expect(capture.length).toBe(2);
         });
@@ -733,7 +732,6 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
                     .published(new CapturingOutputSink(capture))
                     .done()
                     .run(runtime);
-                await Promise.race([sleep(1000), app]);
                 expect(capture.length).toBe(2);
             } catch (err) {
                 error = err;
@@ -783,7 +781,6 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
                     .published(new CapturingOutputSink(capture))
                     .done()
                     .run(runtime);
-                await Promise.race([sleep(1000), app]);
             } catch (err) {
                 error = err;
             } finally {
--- a/packages/core/src/__test__/Application.test.ts
+++ b/packages/core/src/__test__/Application.test.ts
@@ -31,6 +31,7 @@ import {
     MessageProcessingResults,
     MessageRef,
     OutputSinkConsistencyLevel,
+    QueueFullHandlingMode,
     SequenceConflictError,
     StateRef,
 } from "../model";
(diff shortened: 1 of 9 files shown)


packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
```diff
@@ -67,6 +71,35 @@ export class ConcurrentMessageProcessor extends BaseMessageProcessor implements
...
+                                reject(new Error(msg));
+                                break;
+                            case QueueFullHandlingMode.LogAndContinue:
+                            default:
+                                this.logger.warn(msg);
+                                break;
+                        }
+                    }
+                }
+            }, this.config.healthCheck.inputQueueValidation.validationIntervalInMs);
```
plameniv, line 98: Can we use the same value for both validation interval and timeout?

In cases where validation interval is longer than the timeout value, then the timeout value is effectively ignored.

In cases where validation interval is shorter than the timeout value, we perform extra checks which will usually not catch the timeout any sooner.

packages/core/src/model/dsl.ts
```diff
@@ -97,6 +103,15 @@ export interface IParallelismConfiguration {
     readonly concurrencyConfiguration?: IConcurrencyConfiguration;
 }
 
+export interface IQueueValidationConfig {
+    readonly mode: QueueFullHandlingMode;
+    readonly validationIntervalInMs: number;
```
plameniv, line 108: Should we add a default value and what would be a good default value for the validation interval?

packages/core/src/model/dsl.ts
```diff
@@ -97,6 +103,15 @@ export interface IParallelismConfiguration {
...
+    readonly timeOutInMs: number;
+}
+
+export interface IHealthCheckConfig {
+    readonly inputQueueValidation?: IQueueValidationConfig;
+}
```
plameniv, line 114: :nit: Add new line here

packages/core/src/__test__/Application.test.ts
```diff
@@ -653,6 +654,134 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
...
+                    })
+                    .output()
+                    .published(new CapturingOutputSink(capture))
+                    .done()
+                    .run(runtime);
+                await Promise.race([sleep(5000), app]);
+                expect(capture.length).toBe(2);
+            } catch (err) {
+                error = err;
+            }
```
plameniv, line 738: Add `app.cancel` in a finally here. Since the app is never terminated but the race Promise exits and we proceed with the rest of the code.

packages/core/src/__test__/Application.test.ts
```diff
@@ -653,6 +654,134 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
...
+                        onIncrement: async (msg: Increment, ctx: IDispatchContext) => {
+                            await sleep(10);
+                            ctx.publish(Increment, msg);
+                        },
+                    })
+                    .output()
+                    .published(new CapturingOutputSink(capture))
+                    .done()
+                    .run(runtime);
+                await Promise.race([sleep(5000), app]);
```
plameniv, line 734: This sleep can likely be reduced to a smaller value.

packages/core/src/__test__/Application.test.ts
```diff
@@ -653,6 +654,134 @@ for (const mode of [ParallelismMode.Concurrent, ParallelismMode.Rpc]) {
             const published = capture.map((m) => m.message);
             expect(published).toMatchObject([inc(1), inc(2)]);
         });
+
+        xit("successfully validates input queue with length within capacity", async () => {
```
plameniv, line 658: Do we need the `xit` here instead of the `it`? I'd rather keep things consistent.

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
```diff
@@ -67,6 +71,35 @@ export class ConcurrentMessageProcessor extends BaseMessageProcessor implements
...
+                }
+                if (this.inputQueue?.length >= this.config.inputQueueCapacity) {
+                    const currentTimestamp = new Date().getTime();
+                    if (
+                        currentTimestamp - this.lastDispatchedMessageTimestamp >=
+                        this.config.healthCheck.inputQueueValidation.timeOutInMs
+                    ) {
+                        const msg = `Input queue has reached it's capacity, currentLength: ${this.inputQueue.length}, capacity: ${this.config.inputQueueCapacity}`;
+                        switch (this.config.healthCheck.inputQueueValidation.mode) {
+                            case QueueFullHandlingMode.LogAndFail:
```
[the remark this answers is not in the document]
prachi30: I am not implementing retrial at the moment (could be a potential enhancement but I don't see it being required atm) and just logging for retry mode does not sound correct as it would confuse the users.

packages/core/src/__test__/Application.test.ts
prachi30: you caught it while I was testing a theory and of course we shouldn't merge without removing it :)

packages/core/src/model/dsl.ts
prachi30: It would be worth defining a default value if we enable health check by default but since we are just enabling it on request we don't necessarily need it.

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
```diff
@@ -86,16 +118,25 @@ export class ConcurrentMessageProcessor extends BaseMessageProcessor implements
...
+                        serviceDiscovery,
+                        dispatchRetrier
+                    ),
+                    this.outputLoop(sink, inputMessageMetricAnnotator, sinkRetrier),
+                ]
+            );
+
+            await Promise.all(processes);
         } catch (e) {
             this.inputQueue.close();
```
hienqnguyen, line 141: am I reading this right that if and when the validation fails and my service is on `LogAndFail` mode, execution will flow to this `catch` clause, the queue then got shutdown and the service restarts?

hienqnguyen:
It seems to me the nomenclature of `healthCheck` here is a bit confusing.  At first I thought you meant [this](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) but I was wrong.  Maybe `healthCheck` should be reserved for when CC supports it for monitoring in a K8s sense for observability

Also, night be good to add some metrics around this as well

packages/core/src/__test__/Application.test.ts
prachi30: not using sleep any more so closing this out

packages/core/src/internal/processor/ConcurrentMessageProcessor.ts
prachi30: Not necessarily coz timeout is not based on clock but more based on msg processing time

plameniv (approved):
LGTM

PR #375, opened 2022-12-29
Provide longest input mode

YashGajera:
Updated Docs
Screen Shot 2022-12-28 at 10 22 11 PM


plameniv:
If on `true` for this parameter the app only runs until the earliest input is exhausted, shouldn't it be named something other than `longest`?

YashGajera:
> If on `true` for this parameter the app only runs until the earliest input is exhausted, shouldn't it be named something other than `longest`?

It was a mistake in documentation. It is fixed now. If this parameter is true, the app will continue running until all input are exhausted.

docs/docs/Introduction_Inputs.md
```diff
@@ -46,6 +46,21 @@ The lifecycle of a Cookie Cutter application is determined by its input source(s
...
+| -------- | -----| ------- | ----------- |
+| longest  | boolean | false | <b>False</b>: Cookie Cutter application will stop If one input source stops before the others. <br /> <b>True</b>: Cookie Cutter application will keep running until all input sources are exhausted. |
+
+Example:
+```typescript
+Application.create()
+    .input({
+        longest: true
+    })
+        .add(new ExampleInputSource())
```
plameniv, line 61: :nit: indent these 2 lines same way .input is indented

packages/core/src/__test__/internal/helpers.test.ts
```diff
@@ -22,12 +22,36 @@ describe("roundRobinIterators", () => {
...
+        for await (const n of roundRobinIterators(
+            [it1, it2],
+            DefaultComponentContext.logger,
+            true
+        )) {
+            all.push(n);
+        }
+
+        for (const n of [1, 2, 3, 4, 5, 10, 20, 30, 40, 50]) {
+            expect(all).toContain(n);
```
plameniv, line 54: This expect would succeed even if the `all` contained every value from `values1` and `values2`.

Can you add an additional check and add more values in the longer source?

docs/docs/Introduction_Inputs.md
YashGajera: fixed

packages/core/src/__test__/internal/helpers.test.ts
YashGajera: replaced with application test with more values in static input

packages/core/src/__test__/Application.test.ts
```diff
@@ -169,6 +171,82 @@ for (const mode of [ParallelismMode.Serial, ParallelismMode.Concurrent, Parallel
...
+                    onDecrement: async (msg: Decrement) => {
+                        tally -= msg.count;
+                    },
+                })
+                .run(ErrorHandlingMode.LogAndFail, mode);
+            if (mode === ParallelismMode.Serial) {
+                expect(tally).toBe(2);
+            } else {
+                expect(tally).toBe(4);
+            }
```
YashGajera, line 211: @plameniv Concurrent & GRPC mode process messages from queue differently than serial mode?

plameniv: Serial mode processes one message from input step through handler to output step before going to the next message in the input queue.

Concurrent and GRPC add different levels of parallelism in the above process. 

Not yet sure why the tally in the above test has those values.

plameniv (approved):
LGTM

Issue #377, opened 2022-12-29
kafka input source should not be stopped if kafkajs error is retriable by kafkajs 


YashGajera:
kafkajs [CRASH](https://kafka.js.org/docs/instrumentation-events#a-name-list-list-of-available-events) event includes restart flag.
if kafkajs error is retriable, consumer will try to restart itself upon crash and **input source should not be stopped if error is retriable**.

In following example kafkjs crashed with retriable error but input source was stopped immediately reacting to crash event. After that consumer got restarted by kafkajs automatically but it was not consuming any messages because of closed kafka input source pipe. This behavior causes consumer to lag even it is connected to broker and able to consume.

```
{"time":"2022-12-23T03:13:21.374Z","level":"EROR","ctx":"tooling.task-handler.2_0_1","fields":{"msg":["Kafkajs Crashed"],"err":[{"error":{"name":"KafkaJSConnectionError","retriable":true,"broker":"kafka-420262885-13-1196072530.scus.kafka-v2-shared1-prod.ms-df-messaging.prod-az-southcentralus-2.prod.us.walmart.net:9092","code":"ECONNRESET"},"groupId":"prod-chaperone-task-handler-verify-transaction-mis-scans"}]}}
{"level":"ERROR","timestamp":"2022-12-23T03:13:21.374Z","logger":"kafkajs","message":"[Consumer] Restarting the consumer in 300ms","retryTime":300,"groupId":"prod-chaperone-task-handler-verify-transaction-mis-scans"}
{"time":"2022-12-23T03:13:21.375Z","level":"WARN","ctx":"tooling.task-handler.2_0_1","fields":{"msg":["Error iterating input source",{"name":"KafkaJSConnectionError","retriable":true,"broker":"kafka-420262885-13-1196072530.scus.kafka-v2-shared1-prod.ms-df-messaging.prod-az-southcentralus-2.prod.us.walmart.net:9092","code":"ECONNRESET"}]}}
{"level":"INFO","timestamp":"2022-12-23T03:13:21.675Z","logger":"kafkajs","message":"[Consumer] Starting","groupId":"prod-chaperone-task-handler-verify-transaction-mis-scans"}
[...]

PR #378, opened 2022-12-29
Fix for #377: kafka input source should not be stopped if kafkajs error is retriable by kafkajs

YashGajera:
fix for Bug: #377

### DIFF
--- a/packages/kafka/src/KafkaConsumer.ts
+++ b/packages/kafka/src/KafkaConsumer.ts
@@ -247,10 +247,12 @@ export class KafkaConsumer implements IRequireInitialization, IDisposable {
         );
         this.consumer.on(
             this.consumer.events.CRASH,
-            ({ payload: { error, groupId } }: ConsumerCrashEvent) => {
-                this.logger.error("Kafkajs Crashed", { error, groupId });
-                // tslint:disable-next-line:no-floating-promises
-                this.pipe.throw(error);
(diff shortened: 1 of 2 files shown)


packages/kafka/src/KafkaConsumer.ts
```diff
@@ -247,10 +247,12 @@ export class KafkaConsumer implements IRequireInitialization, IDisposable {
         );
         this.consumer.on(
             this.consumer.events.CRASH,
-            ({ payload: { error, groupId } }: ConsumerCrashEvent) => {
-                this.logger.error("Kafkajs Crashed", { error, groupId });
-                // tslint:disable-next-line:no-floating-promises
-                this.pipe.throw(error);
+            ({ payload: { error, groupId, restart } }: ConsumerCrashEvent) => {
+                this.logger.error("Kafkajs Crashed", { error, groupId, restart });
+                if (!restart) {
```
prachi30, line 252: nit: worth adding some unit tests for this

YashGajera: I don't think there is a way to emit kafkajs crash event from jest. I tried mocking node events module but that did not worked because eventemitter instance gets created in kafakjs and its not exposed outside of kafkajs. we can mock the whole kafkajs module but that will be too much work just to test handling of crash event. Let me know if you have any suggestions or experience with event emitter mocking in jest.

PR #392, opened 2024-02-13
add orderingKey for pubSub sink

amitlopes-walmart:


packages/gcp/src/PubSubSink.ts
```diff
@@ -33,10 +33,17 @@ interface IPayloadWithAttributes {
...
+}
+
+interface IPubSubTopicPayload {
+    payloadWithAttributes: IPayloadWithAttributes[];
+    messageOrdering: boolean;
 }
 
 export enum PubSubMetadata {
     Topic = "topic",
+    OrderingKey = "orderingKey",
```
butorine: I would rename this to `Key = "orderingKey"` to be consistent with kafka nomenclature

packages/gcp/src/PubSubSink.ts
```diff
@@ -70,15 +77,21 @@ export class PubSubSink
...
+                    payloadWithAttributes: [],
+                });
             }
 
             const formattedMsg = this.formatMessage(msg);
-            messagesByTopic.get(topic).push(formattedMsg);
+            const topic = messagesByTopic.get(topicName);
+
+            topic.payloadWithAttributes.push(formattedMsg);
+            topic.messageOrdering = !!formattedMsg.orderingKey;
```
butorine: This should only be set if `topic.messageOrdering` is false.  If it's already true, then bypass.  This way if some messages contain a key, but some do not, then all will be published with an ordering key

packages/gcp/src/PubSubSink.ts
```diff
@@ -33,10 +33,17 @@ interface IPayloadWithAttributes {
     payload: Buffer;
     attributes: Attributes;
     spanContext: SpanContext;
+    orderingKey: string;
+}
+
+interface IPubSubTopicPayload {
+    payloadWithAttributes: IPayloadWithAttributes[];
```
butorine: Make this plural, please: `payloadsWithAttributes`

butorine (changes requested):

packages/gcp/src/__test__/PubSubSink.test.ts
```diff
@@ -139,12 +144,88 @@ describe("PubSubSink Tests", () => {
...
-            expect(mockPublishFn.mock.calls[idx][1][AttributeNames.contentType]).toBe(
+            expect(mockPublishFn.mock.calls[idx][0].attributes[AttributeNames.eventType]).toBe(
+                message.type
+            );
+            expect(mockPublishFn.mock.calls[idx][0].attributes[AttributeNames.contentType]).toBe(
                 pubSubPublisherConfigurationWithDefaultTopic.encoder.mimeType
             );
         });
     });
+    it("with orderingKey", async () => {
```
plameniv, line 158: Can you make sure to have an empty line before each `it(...` line?

packages/gcp/src/__test__/PubSubSink.test.ts
```diff
@@ -139,12 +144,88 @@ describe("PubSubSink Tests", () => {
...
+            expect(mockTopic.mock.calls[idx][1].messageOrdering).toBeTruthy();
+        });
+        expect(mockPublishFn).toBeCalledTimes(messagesWithTopic.length);
+        messagesWithTopic.forEach((message, idx) => {
+            expect(mockPublishFn.mock.calls[idx][0][PubSubMetadata.Key]).toBe(
+                message.payload.orderingKey
+            );
+        });
+    });
+    it("without orderingKey", async () => {
```
plameniv: I think you can just move a couple of the expects which check for `no orderingKey` into the 2 already existing positive tests which don't use an ordering key.
And remove this `it(...` segment.

plameniv (approved):
LGTM!

PR #394, opened 2024-05-29
Add API key to gRPC server and client

plameniv:


packages/grpc/src/internal/client.ts
```diff
@@ -86,7 +88,7 @@ export function createGrpcClient<T>(
 
         const metaCallback = (_params: any, callback: (arg0: null, arg1: Metadata) => void) => {
             const meta = new Metadata();
-            meta.add("custom-auth-header", "token");
+            meta.add("custom-auth-header", apiKey || DEFAULT_API_KEY);
```
YashGajera: I think we can skip adding the header if there is no API Key provided. Also header name should be "authorization"
```suggestion
if(apiKey) {
            meta.add("authorization", apiKey);
            }
```

packages/grpc/src/internal/client.ts
```diff
@@ -75,7 +76,8 @@ class ClientBase implements IRequireInitialization, IDisposable {
 
 export function createGrpcClient<T>(
     config: IGrpcClientConfiguration & IGrpcConfiguration,
-    certPath?: string
+    certPath?: string,
+    apiKey?: string
```
YashGajera: May be we can turn this into object while this version is still in beta.
option?: {
```suggestion
options?:  {  certPath?: string,
    apiKey?: string
    }
```

packages/grpc/src/internal/client.ts
```diff
@@ -41,6 +41,7 @@ enum GrpcMetricResult {
     Success = "success",
     Error = "error",
 }
+const DEFAULT_API_KEY = "token";
```
YashGajera: I don't think we need this

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -239,4 +257,12 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
             message: error.toString(),
         };
     }
+
+    private isApiKeyValid(meta: Metadata) {
+        const headerValue = meta.get("custom-auth-header");
```
YashGajera:
```suggestion
        const headerValue = meta.get("authorization");
```

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -239,4 +257,12 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
...
+        if (!headerValue || headerValue.length < 1 || headerValue[0].toString() !== this.apiKey) {
```
YashGajera, line 263:
```suggestion
        if (headerValue?.[0]?.toString() !== this.apiKey) {
```

packages/grpc/src/internal/GrpcInputSource.ts
```diff
@@ -239,4 +257,12 @@ export class GrpcInputSource implements IInputSource, IRequireInitialization {
...
+            return false;
+        }
+        return true;
```
YashGajera:
```suggestion
        return headerValue?.[0]?.toString() === this.apiKey;
```

plameniv: The suggested correction does not handle an empty array for the headerValue. Do you think that's a unlikely case?

YashGajera: It should handle empty array.
Screenshot 2024-05-29 at 2 17 14 PM

packages/grpc/src/internal/client.ts
plameniv: Added the if check a bit earlier

packages/grpc/src/internal/client.ts
plameniv: Created a separate interfaces for server and client

YashGajera (approved):

packages/grpc/src/index.ts
```diff
@@ -102,7 +112,7 @@ export function grpcMsg(operation: IGrpcServiceMethod, request: any): IMessage {
 
 export function grpcClient<T>(
     configuration: IGrpcClientConfiguration & IGrpcConfiguration,
-    certPath?: string
+    options?: IGrpcClientOptions
```
cross311: This change is a breaking change.  Can we do it in a non breaking way?  Allow cert to still be passed in and add options and Che k both.

cross311 (commented):
Either update version to be a breaking change version or just allow support for old and new way.

talam (approved):

PR #395, opened 2025-01-17
BUG FIX: Only quit redis client if it's ready

YashGajera:
Invoking quit() on a Redis client that is not ready causes the application to hung, and the CC App is unable to dispose of the component, resulting in a hung CC application.

### DIFF
--- a/packages/redis/src/RedisClient.ts
+++ b/packages/redis/src/RedisClient.ts
@@ -207,7 +207,9 @@ export class RedisClient implements IRedisClient, IRequireInitialization, IDispo
 
     public async dispose(): Promise<void> {
         this.disposeInitiated = true;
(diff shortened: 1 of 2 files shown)


packages/redis/src/RedisClient.ts
```diff
@@ -207,7 +207,9 @@ export class RedisClient implements IRedisClient, IRequireInitialization, IDispo
 
     public async dispose(): Promise<void> {
         this.disposeInitiated = true;
-        await this.client.quit();
+        if (this.client.isReady) {
```
cross311: I would add a comment above this as to why only calling quit when he client is actually connected.

cross311 (approved):

cross311 (approved):

PR #396, opened 2025-03-27
Update Redis client config with new options for username, ca & TLS

omkrk:


packages/redis/src/RedisClient.ts
```diff
@@ -34,7 +34,6 @@ import {
     WatchError,
 } from "redis";
 
-import { isNullOrUndefined } from "util";
```
omkrk, line 37: Updating this function as it is marked as deprecated since v4.0.0 node version

cross311 (approved):

PR #400, opened 2025-05-22
Fix Redis Stream Acknowledgment Error That Causes Application Deadlock

### LINKED ISSUE #399: Redis Stream Acknowledgment Error Causes Application Deadlock
The application can get stuck in an infinite loop when Redis acknowledgments (`xAck`) fail. This happens because the message processing pipeline does not properly handle errors during the message acknowledgment phase, preventing messages from being removed from the pending set in `CompositeInputSource`.

### Potential Causes of xAck Failures

1. Message already acknowledged by another process
2. Redis server restarts or failovers
3. Redis cluster rebalancing
4. Configuration issues with streams/consumer groups
5. Network issues between application and Redis EOF

## Impact

When this issue occurs:
1. The application becomes unresponsive
2. No new messages are processed
3. The application cannot shut down gracefully

### Message Flow
1. Messages are received from Redis Streams
2. Messages are processed by the application
3. Upon completion, messages should be acknowledged to Redis
4. Message references should be removed from the pending set

### Error Scenario
When a Redis `xAck` operation fails (returns 0 or throws an exception), the error handling in `RedisStreamSource.onMessageReleased` throws an exception:

```typescript
// In RedisStreamSource.ts:
private async onMessageReleased(span: Span, msg: MessageRef, err: Error) {
    const stream = msg.metadata<string>(RedisStreamMetadata.Stream);
    const messageId = msg.metadata<string>(RedisStreamMetadata.MessageId);
    const consumerId = msg.metadata<string>(RedisStreamMetadata.ConsumerId);

    try {
        if (err) throw err;

        const count = await this.client.xAck(span.context(), stream, consumerId, messageId);
        if (count !== 1) {
            throw new Error("not found in PEL"); // This error propagates and breaks the flow
        }

        // Success metrics...
    } catch (e) {
        // Error metrics...

        if (!err) {
            this.logger.error("failed to ack message", e, { messageId, stream, consumerId });
            throw e; // This error propagates and breaks the flow
        }
    } finally {
        span.finish();
    }
}
```
This error prevents the message from being properly removed from the pending set in CompositeInputSource:
```typescript
// In CompositeInputSource.ts:
pending.add(msg);
msg.once("released", async (m) => {
    pending.delete(m); // This never happens if an error is thrown in onMessageReleased
});
```
Leading to the application getting stuck in the infinite loop:
```typescript
// In CompositeInputSource.ts:
// only terminate this generator after all
// elements yielded have been released
while (pending.size > 0) {
    await sleep(100);
    await waitForPendingIO();
}
```

YashGajera:
### Problem
Applications using Redis streams can get stuck in an infinite loop when Redis acknowledgments ([xAck](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)) fail. This happens because the message processing pipeline does not properly handle errors during the message acknowledgment phase, preventing messages from being removed from the pending set in CompositeInputSource.

When an [xAck](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) operation returns 0 or throws an exception, the error handling in [RedisStreamSource.onMessageReleased](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) throws an exception, breaking the message release flow. This prevents messages from being properly removed from the pending messages set, causing the application to get stuck in the waiting loop in CompositeInputSource.

### Solution
This PR modifies the error handling in [RedisStreamSource.onMessageReleased](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) to log errors instead of throwing exceptions when acknowledgment fails. This ensures the message release flow completes regardless of acknowledgment failures, preventing the application from deadlocking.

## Key changes:

Instead of throwing an error when [count !== 1](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), we now log a warning message
In the catch block, we log errors but don't propagate them when !err is true
Added more detailed logging to help with troubleshooting
### Testing
Verified that the application continues processing messages even when Redis acknowledgments fail
Confirmed that errors are properly logged for troubleshooting
Checked that metrics are still updated to reflect acknowledgment failures
### Related Issues
Fixes #399

### DIFF
--- a/packages/redis/src/RedisStreamSource.ts
+++ b/packages/redis/src/RedisStreamSource.ts
@@ -238,9 +238,8 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization, 
             if (count !== 1) {
                 // Log the error but don't throw, to prevent blocking the message release process
                 // This approach prioritizes system stability (preventing deadlocks) over strict message processing guarantees. In a distributed system, this is often a reasonable tradeoff, as you can address message acknowledgment issues through monitoring and alerts rather than causing the entire application to hang.
-                this.logger.error(`Message ack returned ${count} (expected 1), message likely not in PEL`,
-                    { messageId, stream, consumerId });
-                this.logger.error("failed to ack message", { messageId, stream, consumerId });
+                this.logger.error("Message ack returned (expected 1), message likely not in PEL",
+                    { messageId, stream, consumerId, xAckCount: count });
                 this.metrics.increment(RedisMetrics.MsgAcknowledged, {
                     stream_name: stream,
                     consumer_group: consumerId,
--- a/packages/redis/src/RedisStreamSource.ts
+++ b/packages/redis/src/RedisStreamSource.ts
@@ -29,6 +29,7 @@ export enum RedisMetrics {
     MsgsClaimed = "cookie_cutter.redis_consumer.input_msgs_claimed",
     PendingMsgSize = "cookie_cutter.redis_consumer.pending_msg_size",
     IncomingBatchSize = "cookie_cutter.redis_consumer.incoming_batch_size",
+    MsgAcknowledged = "cookie_cutter.redis_consumer.input_msg_acknowledged",
 }
 
 export enum RedisMetricResult {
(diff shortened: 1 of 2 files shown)


packages/redis/src/RedisStreamSource.ts
```diff
@@ -235,7 +236,16 @@ export class RedisStreamSource implements IInputSource, IRequireInitialization,
 
             const count = await this.client.xAck(span.context(), stream, consumerId, messageId);
             if (count !== 1) {
-                throw new Error("not found in PEL");
+                // Log the error but don't throw, to prevent blocking the message release process
+                // This approach prioritizes system stability (preventing deadlocks) over strict message processing guarantees. In a distributed system, this is often a reasonable tradeoff, as you can address message acknowledgment issues through monitoring and alerts rather than causing the entire application to hang.
+                this.logger.error(`Message ack returned ${count} (expected 1), message likely not in PEL`,
+                    { messageId, stream, consumerId });
+                this.logger.error("failed to ack message", { messageId, stream, consumerId });
```
cross311: Two log messages where they both say the same thing. Can we remove the line 243.

cross311 (changes requested):
double log message.

packages/redis/src/RedisStreamSource.ts
YashGajera: removed

cross311 (approved):

PR #405, opened 2026-03-24
Add timeout

b0b0ddq:
In gcp package, when uploading data to google pubsub 'batchPublisher.publishMessage' makes a grpc call internally. Currently there is no timeout functionality to this call and if one call hangs, it blocks the whole sink processing. This issue is causing transaction-uploader-pubsub service to hang until the service is restarted. 

This change is to add timeout on  publishMessage call and add metrics to track upload times .

### DIFF
--- a/packages/gcp/src/PubSubSink.ts
+++ b/packages/gcp/src/PubSubSink.ts
@@ -140,6 +140,7 @@ export class PubSubSink
                             topic,
                             eventType,
                         });
+                        throw e;
                     }
                 } finally {
                     span.finish();
--- a/packages/gcp/src/PubSubSink.ts
+++ b/packages/gcp/src/PubSubSink.ts
@@ -127,7 +127,13 @@ export class PubSubSink
                     this.logger.debug("Message published to PubSub", { topic, messageId });
                 } catch (e) {
                     failSpan(span, e);
-                    this.emitMetrics(topic, eventType, PubSubMetricResults.Error);
+                    topicPayload.messages.forEach((message) =>
+                        this.emitMetrics(
+                            topic,
+                            message.attributes[AttributeNames.eventType],
+                            PubSubMetricResults.Error
+                        )
+                    );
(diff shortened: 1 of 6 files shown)


packages/gcp/src/index.ts
```diff
@@ -134,6 +135,7 @@ export function pubSubSink(
         maximumBatchSize: 1000,
         maximumBatchWaitTime: 100,
         maxPayloadSize: 5242880,
+        publishTimeoutMs: 0,
```
talam: the default timeout here needs to be longer. ideally it should match whatever the underlying gcp sdk uses/recommends.

packages/gcp/src/PubSubSink.ts
```diff
@@ -170,6 +179,43 @@ export class PubSubSink
...
     }
 
+    private emitPublishTimingMetric(topic: string, eventType: string, runTime: number) {
+        this.metrics.timing(PubSubMetrics.PublishTime, runTime, {
+            topic,
+            event_type: eventType,
+        });
+    }
+
+    private async publishMessageWithTimeout(
```
talam: i don't think this is the correct way to do this. the pubsub client supports timeouts as part of its config.
https://googleapis.dev/nodejs/pubsub/latest/Topic.html#setPublishOptions
https://googleapis.dev/nodejs/pubsub/latest/global.html#PublishOptions
https://googleapis.github.io/gax-nodejs/interfaces/CallOptions.html

there's a gax option called timeout. should we be using that instead?

talam (changes requested):
I don't think we need to manually setup timeouts. Take a look at the sdk's configuration options. It should be natively supported.

packages/gcp/src/PubSubSink.ts
b0b0ddq: good one. Using native timeout instead of custom implementation

talam (approved):

